PR #288, opened 2022-08-11
Fix missing comma causing install failure

darkorb:
L131 has a missing `,` at the end resulting in this not being valid for parsing.

This simply adds it, allowing new installs to work again 😸


aaronwmorris:
Sigh.  Thanks for catching this!

PR #336, opened 2022-08-30
Add support for external temperature script for cameras without temperature sensors

aaronwmorris:


jkoenig72:
Thx - will try it out as soon I get some free time.

PR #436, opened 2022-11-26
mod: passed setup.sh through shellcheck and resolved warnings

aaronwmorris:


aaronwmorris:
There were a couple of fixes that were needed.  I also fixed the remaining shellcheck issues that were popping up.

PR #450, opened 2022-12-11
ffmpeg webm support

Elliot-Ford:
I've added the option to change the output codec for the timelapse. Webm has more compatibility across browsers (I've had issues with x264) and is a lower bitrate codec than x264. It does take longer to generate the resulting images (I've had one timelapse take 4hrs) but I didn't run out of memory on my 2gb pi in tests where the x264 codec did.

Logging for ffmpeg also had to be adjusted as default logging level with webm caused it too fail due to exceeding max log message length.


aaronwmorris:
Cool, I'll take a look at this after I finish with the other PR.

PR #790, opened 2023-06-04
Add date options to remote path

gshau:
This PR adds two additional options to the remote path to help facilitate more organized upload paths
- `date` - gives the date string of the current time
- `night_date` - gives the date corresponding to the beginning of the night

### DIFF
--- a/indi_allsky/flask/forms.py
+++ b/indi_allsky/flask/forms.py
@@ -1324,6 +1324,8 @@ def FILETRANSFER__REMOTE_IMAGE_NAME_validator(form, field):
         'timestamp'  : datetime.now(),
         'ts'         : datetime.now(),
         'ext'        : 'jpg',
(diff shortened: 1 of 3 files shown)


aaronwmorris:
I really appreciate the pull request, although what you have added is [mostly] already in place.

You can already template the current date stamp in the remote folders using variable replacement.  In the remote path you can use the timestamp/ts variables directly like this:

`/remote/path/{timestamp:%Y-%m-%d}/`

What I do like about what you added is the timestamp for the current night, which I do not have.

I will merge the request so you get added to the contributors, but I will modify the functions to remove what is already implemented.

gshau:
Nice, didn't think to use the format specifier for date.  Good to know!  Thanks for the merge!

aaronwmorris:
I completed the merge in #792

PR #834, opened 2023-07-12
Implemented time estimation in the dark exposures loop

interplanetarychris:
- Added function `_estimate_runtime` to estimate the remaining runtime of the `_run` function.
- Added variable `remaining_configs` to track the number of remaining configurations.
- Added variables `total_exposures`, `overhead_per_exposure`, and `completed_exposures` to track the total number of exposures, the overhead time per exposure, and the number of completed exposures, respectively.
- Added function `format_time` to format an integer number of seconds as HH:MM:SS. Accelerated daytime darks by limiting to a single 1 second * count exposure.

### DIFF
--- a/indi_allsky/darks.py
+++ b/indi_allsky/darks.py
@@ -569,6 +569,26 @@ class IndiAllSkyDarks(object):
                 self.indiclient.configureCcdDevice(self.config['INDI_CONFIG_DEFAULTS'])
 
 
+    @staticmethod
+    def _format_time(seconds):
+        """Take an integer number of seconds and return a string in the format HH:MM:SS."""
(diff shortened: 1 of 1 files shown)


aaronwmorris:
Good stuff.  Let me review this.

I am about to submit a rather large change that refactors the capture code for [light] images, not darks.  This should not interfere with that.

aaronwmorris:
I am a little irritated GitHub has not added you to the contributors yet.  I may just create a file in the repo.

interplanetarychris:
Don’t worry about it - will probably have some more pull requests at some point.

-Chris
@interplanetary

PR #911, opened 2023-08-24
fix smoke rating when legacy string present

aaronwmorris:
New image column to exclude images from timelapse and startrails


Vertikar:
Nice work fixing this one quickly! I ran into it yesterday

PR #1226, opened 2024-04-20
Corrected name of base image docker_indi.allsky.base

Mats-Code-Journey:
Name of base image adapted to make docker-compose build the images correctly

### DIFF
--- a/docker/Dockerfile.capture
+++ b/docker/Dockerfile.capture
@@ -1,6 +1,6 @@
 
 # docker- prefix due to files in docker/ folder
-FROM docker-indi.allsky.base
+FROM docker_indi.allsky.base
 
 ARG DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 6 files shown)


aaronwmorris:
Is it possible that you used `docker-compose` instead of `docker compose` (no dash).  On Linux, `docker-compose` defaults to the docker compose v1, but `docker compose` uses the compose v2 functionality.

Mats-Code-Journey:
That‘s correct. `docker compose` gives an error to me, `compose is not a docker command`, However `docker-compose build` did start correctly but had the problem to build the base image due to the `-` instead of `_`.
All is run on a fresh install of Raspberry Pi 5 with default package repositories. It seems that `docker compose ` can be added with a docker install from the docker repositories.

After some more research I am now aware of the differences of `docker compose ` and `docker-compose`. 
I will try again without the change.

Mats-Code-Journey:
With 'docker compose' installed from docker repositories it builds fine. After I have found the comments in the docker-compose.yml also libcamera works now.

PR #1422, opened 2024-07-25
ensure S3 key is not a Pathlib object

aaronwmorris:


gaitskell:
@aaronwmorris Thanks for doing it so quickly - this worked well.

PR #1622, opened 2024-11-08
docker: gunicorn: base on python:3.11 and shave 4gb off the image

USA-RedDragon:
This patch reduces the final size of the gunicorn image from a whopping 5.4GB down to 1.4GB. The Python dependencies alone take up 1.1GB (`/usr/local/lib/python3.11/site-packages`).

### DIFF
--- a/docker/Dockerfile.gunicorn
+++ b/docker/Dockerfile.gunicorn
@@ -1,13 +1,78 @@
 
 # docker- prefix due to files in docker/ folder
-FROM docker-indi.allsky.base
+FROM python:3.11-slim
 
 ARG DEBIAN_FRONTEND=noninteractive
 
 USER root
-RUN apt-get update
-RUN apt-get -y upgrade
 
+RUN apt-get update \
+    && apt-get -y upgrade \
+    && apt-get -y install \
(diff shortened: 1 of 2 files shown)


aaronwmorris:
I think this image could be made even more simple.  indilib has not been a requirement for the gunicorn process for a while, I just never revisited this particular container.

libindi and pyindi-client should no longer be necessary.

USA-RedDragon:
Not a huge impact, less than 100mb so `dive` is still reporting 1.4gb

aaronwmorris:
One last change here... I have a reduced set of python requirements for the gunicorn image:  `requirements_latest_web.txt`

aaronwmorris:
Looks good to me, I will merge if you are ready.

USA-RedDragon:
Using the gunicorn-specific requirements brings the total image size to a nice round 1.0GB. This one should be good to ship

USA-RedDragon:
Wait on this one, seems there's a missing dep

USA-RedDragon:
Ok, tested with the patch above and it fires up properly now. This one should be good to go

PR #1623, opened 2024-11-09
feat(boto_minio): add minio client for object storage

USA-RedDragon:
In order to use Minio, I needed to set a custom endpoint. I noticed this was commented out, so I went ahead and wired it up.

### DIFF
--- a/indi_allsky/filetransfer/__init__.py
+++ b/indi_allsky/filetransfer/__init__.py
@@ -15,6 +15,7 @@ from .pycurl_webdav_https import pycurl_webdav_https
 from .paho_mqtt import paho_mqtt
 
 from .boto3_s3 import boto3_s3
+from .boto3_minio import boto3_minio
 from .libcloud_s3 import libcloud_s3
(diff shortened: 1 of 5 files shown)


USA-RedDragon:
Converting to draft. For the upload, the `endpoint_url` shouldn't contain the bucket name in the path, but the capture process sets the s3_prefix properly, so this patchset breaks uploads

USA-RedDragon:
Fixed with 8c45e2239128bdfde1b136e1bc2b4bf9bd3fa2e3

aaronwmorris:
What do you think about implementing this as a custom `S3 Utility` in the drop down and keep the origin AWS boto3 code as-is?

USA-RedDragon:
I've got zero issue with that, I'll pursue that direction. :)

aaronwmorris:
Looks good to me, I will merge if you are ready.

USA-RedDragon:
Just validated these changes, this one is good

PR #1624, opened 2024-11-09
More Docker minifications

USA-RedDragon:
Capture container drops from 5.4GB to 1.8GB
INDI base container drops from 3.4GB to 2.3GB


aaronwmorris:
Is the TZ variable not needed?  I am assuming the container automatically inherits the TZ from the host?

USA-RedDragon:
Ahh, so typically (LinuxServer.io containers are a good example of this), Docker images ship as UTC, then users are expected to put a `TZ` environment variable with their timezone rather than hard-coding a TZ at build time if they want the app's time to be localized. As long as the `tzdata` is in the container, a runtime `TZ` env is all that's needed.

aaronwmorris:
If you want to merge the ci.yaml I can definitely work on getting the GitHub actions setup.  I may have to reach out for help since I do not have a great amount of familiarity with this.

PR #1791, opened 2025-01-21
Update requirements_latest_web.txt

fferrann:
## Description
When trying to run the web server only via docker compose, the following error is shown:
error
The application fails to run as the container keeps restarting in a loop.

## Proposed Solution
Add the missing dependency

## Testing
I tested it by:
- Adding the missing dependency 
- Re-building the containers
- Running docker compose up

### DIFF
--- a/requirements/requirements_latest_web.txt
+++ b/requirements/requirements_latest_web.txt
@@ -41,4 +41,4 @@ requests[security]
 lxml
 shapely >= 2.0.6
 pytz
-mysql-connector
+mysql-connector-python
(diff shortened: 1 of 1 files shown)


aaronwmorris:
I recently moved the mysql module to be a primary module and completely forgot to add it to the web requirements.  Thanks for catching this.

Can you change the module from `mysql-connector` to `mysql-connector-python`.  The former is deprecated.

fferrann:
I made the change as suggested and re-tested succesfully.

PR #1866, opened 2025-02-09
Ecowitt API support

aaronwmorris:


WirthmU:
I have added EcoWitt API and KEY and device from my GW300A weather station but only zero values in the graphs.
Is there a way I could check a log file to see why this is not working?

aaronwmorris:
@WirthmU You can use the `./misc/sensor_test.py` script to see if the API is throwing any errors.

PR #1950, opened 2025-03-02
Fix Celsius typo

fferrann:
There is a typo in the word 'Celcius'. It should be Celsius. source: https://en.wikipedia.org/wiki/Celsius


aaronwmorris:
image

PR #2034, opened 2025-03-31
Add include/exclude function to gallery

aaronwmorris:


indi_allsky/flask/templates/gallery.html
```diff
@@ -163,6 +163,33 @@
...
+function exclude_image(el, pswp, image_id, exclude) {
+    var json_data = {
+        'CAMERA_ID'        : fields["CAMERA_ID"].input.val(),
+        'EXCLUDE_IMAGE_ID' : image_id,
+        'EXCLUDE_EXCLUDE'  : exclude,
+    };
+
+    $.ajax({
+        type: "POST",
+        url: "{{ url_for('indi_allsky.ajax_image_exclude_view') }}",
```
omeganon, line 175: Where are these URLs defined? I did a cursory search yesterday for the detections one and I didn’t find a corresponding definition in the codebase.

aaronwmorris: Most are in `indi_allsky/flask/views.py`.  The "indi_allsky" prefix is the name of the flask blueprint.  The "ajax_image_exclude_view" is the name of the route which will be near the bottom of the view.py file.

PR #2296, opened 2025-08-10
Fix typo in conf files

campbellbartlett:
I think this might be a typo? The property that these comments refer to are in `flask.json`. I was a little confused at first, looking for the flash config 😂


aaronwmorris:
Indeed.  My brain said one thing, but my fingers decided on something else.  :-)

PR #2596, opened 2025-12-29
Add sign to threshold settings

aaronwmorris:
A few other settings were modified to include signs, as well.


danteffm:
Works perfectly!

PR #2632, opened 2026-01-14
Added some features helpful for mobile use

Rob0xFF:


aaronwmorris:
Very interesting with the UPS Hat!

I had a question.  Would it be possible to attempt to use the Circuit Python interface for I2C on the UPS hat in the indi-allsky code (not necessarily the watchdog)?  One thing about using Circuit Python instead of SMBus is that it allows you to access multiple devices on the same I2C bus from separate modules because it uses a singleton for the I2C access.

Rob0xFF:
Yes – that should be doable.
The UPS HAT MCU is accessed via I²C address 0x2D with a simple “register pointer + block read” protocol. We can implement that using CircuitPython/Blinka (board.I2C() + adafruit_bus_device.i2c_device.I2CDevice) instead of smbus2, so the UPS monitor shares the same I²C singleton with other devices (BME280/BH1750, etc.) within the same process. I will try to change the implementation as soon as possible. 

Should we implement the USP hat as another sensor, or would it be better to implement it as "device"?

Thanks, Rob

aaronwmorris:
I think a sensor is fine.  We could even implement the watchdog feature directly in the sensor code.  I have some pre-existing code to shutdown via DBus programmatically.

aaronwmorris:
Looks good to me.  The only thing left is to rename the module alias `ups_hat_waveshare_e_mcu_i2c` to `blinka_ups_hat_waveshare_e_mcu_i2c`.

The `blinka_` prefix just allows it to process the flask form validator to ensure the modules are installed and the system can utilize it.  https://github.com/aaronwmorris/indi-allsky/blob/main/indi_allsky/flask/forms.py#L5607

I can pull this in my dev branch if you want, or you can take care of it.  It is a really minor change.

Rob0xFF:
Thanks for the review and for pointing out the blinka_ prefix requirement in the form validator.

Happy to take you up on your offer — please go ahead and pull the change into your dev branch on your side. That’s totally fine with me.

One additional thought (separate from this UPS change): it would be really useful if indi-allsky could support more than the current four configurable sensors (SENSOR A–D). With setups like BME280 + BH1750 + UPS + extra probes, it’s easy to outgrow four slots. If you think that’s something you’d consider, I’m happy to help with an approach/PR (either extending to A–F as a small step, or refactoring to a list-based configuration longer term).

aaronwmorris:
Merged to main.

I would not be opposed to a couple more sensors.  I can work on this, I will probably take the shortest path and just copy/paste the additional sensors.

PR #2688, opened 2026-02-04
Add Download button for VirtualSky image

omeganon:
This commit uses html2canvas (https://html2canvas.hertzen.com) to screenshot the image container and auto-download the png image.

You'll probably want to  change the location of the download button. UI/UX is not my strong suit.

UI --

Screenshot 2026-02-04 at 10 50 54 AM

One caveat is that the downloaded image will be the same size as the displayed image in the browser since it is just a screenshot type of capture.

### DIFF
--- a/indi_allsky/flask/templates/virtualsky.html
+++ b/indi_allsky/flask/templates/virtualsky.html
@@ -578,11 +578,9 @@ $( document ).ready(function() {
 
       if (document.fonts) await document.fonts.ready;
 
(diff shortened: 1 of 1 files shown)


omeganon:
example downloaded image (at 96 DPI) --

virtualsky

omeganon:
html2canvas does support specifying the canvas size height/width to render. If that information for the allsky image can be passed to the function then that would ensure that the captured image sized matched the original source allsky image - https://html2canvas.hertzen.com/configuration

PR #2727, opened 2026-02-23
Experiment with sharpening

ikeysolomon:
Hi Aaron, 

I have had a go at doing a PR for sharpening. It is mainly AI generated but it works. I have nested this with along with existing Saturation and Gamma under the Image tab. If you max it out at 2.0 I believe you will definitely be able to see that it works compared to not. To my eye it makes the image pop more, adds clarity and could compensate a little for poorer focus. 

Open to any and all feedback you might have for me. 

Kind Regards,
Sam

### DIFF
--- a/docker/env_template
+++ b/docker/env_template
@@ -5,6 +5,12 @@
 TZ=America/New_York
 
 INDIALLSKY_INDI_CCD_DRIVER=indi_simulator_ccd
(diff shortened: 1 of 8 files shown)


ikeysolomon:
testing with a page of text bluetacked on a wall in an office, you can definitely see the text becomes much more readable.

PR #2730, opened 2026-02-23
Add image sharpening function

aaronwmorris:
Thank you @ikeysolomon


ikeysolomon:
No, thank you for the opportunity.

aaronwmorris:
I also had to merge #2732 so the sharpen option worked on the image processor.

PR #2759, opened 2026-03-05
Denoising module for consideration

ikeysolomon:
Hi Aaron,

This pull request adds a new lightweight denoising subsystem for the indi-allsky pipeline. It’s fully functional and has been exercised on my back‑yard data, but I haven’t tested anywhere else. So, I’m
passing it on for broader testing and review. The patch is fairly large ~1000 LoC, so no rush. Further, some of the tunable parameters might need some push-pull to bed in first. 

Please feel free to chop, change and do whatever to refine this. Nothing is off the table.

## What’s included

I have chosen to model 4 algorithms (Is that too many? Everything is negotiable here). I have done so with the following points in mind:

- Try for low-cost compute, even at highest quality.
- The ability to cover a range of noise types.
- A range of options for users with Rpi's starting at 3b say up to 5.
- Can be implemented within a pipeline and achieved in quick timeframe.

## Models and **real world** processing time (night):

- Bilateral filter 			~4s processing
- Gaussian blur filter 			~4s processing
- Median blur filter 			~4s processing
- Wavelet filter 	(Daubechies wavelet)	~5s processing (Slowest but highest quality)

**On my rpi5 with 8gb ram and IMX178** saving as level 2 compression *.png. Of note the resultant images end up smaller in mg on disk due to the removal of the noise but retain the same resolution. 

I also tried a non-local-means (NLM) method, but this proved way too costly on compute processing time, also a hot pixel detection method but this competed too much with the star detection in the first place. Just have to rely on darks. 

## New Dependencies
The pywt package is now needed as a module to be installed for the wavelet denoise, but it's only like 5 meg.
The photoutils package is now need as a module for DOAStarFinder (for the star mask). It works very very well, I am a big fan. 

## Particulars
The strength of the denoising algorithms are tunable so hopefully a user should be able to find a happy ground with a particular algorithm and strength that suits them?

In the pipeline this denoising occurs after dark subtraction but before stretching, as this makes sense to not stretch more noise 'into' the image but rather remove it first. The logging looks a little counterintuitive because the dark subtraction appears beforehand, but the timestamp is correct order. 

In order not to scrub out stars whilst denoising I have created a protection_masks module within this work. It computes a star mask.

The star mask uses DAOStarFinder (photutils) to detect point sources and stamps a Gaussian PSF at each detected star. The result is a binary circle which protects stars, it can be dialed up or down using a variable to tune. An LRU cache keyed on an MD5 digest of the frame avoids recomputing the mask when the same frame is seen again. The star mask is now tuneable (DENOISE_STAR_SIGMA, DENOISE_STAR_FWHM, DENOISE_STAR_PERCENTILE) — an earlier version bombed on compute time, but switching to sigma-clipped background stats + the PSF stamp approach brought it well within budget.

The Star Mask itself takes around 1 sec to run, incorporated into the times already listed above. I have a time gate on this so that the star mask is skipped between 5am and 5pm (not during the day) because unneeded. Further, there is print logging logic to track when the period of exposure or interval time is shorter than 5 sec. People who set this low may have trouble with denoising, but it will become obvious in the log. 

There is luminance matching so that the denoising itself doesn't return a duller image because of the choice to use denoising. 

I tried a Milky Way mask but this way too hard and bombed on compute time. It will be on the user not to overdo the denoising. 

## DENOISE PR TEST ENVIRONMENT DIRECTORY
I've included a test script (test_pr_real_image.py) that generates visual output show the star masks on a real backyard image. Might be handy for review — the overlay makes it pretty clear what's being protected and what's being denoised. The script runs at max strength and produces labeled images of the different methods all at a strength of level 5. 

## Some things worth flagging for your consideration:
- The work has not been integrated fully into processing.py but just points outside and calls in: denoising.py which subsequently calls protection_masks.py. 
- photutils and pywavlets are now hard dependencies for this work.

I am grateful for any and all feedback you might have for me. Given that I have leveraged AI heavily to achieve this, I might suggest multiple reviewers if you have any interested parties at hand?

Apologies for the mouthful. I am only new to development and have leveraged AI hugely. 
Thanks again,
Kind Regards,

Sam

### DIFF
--- a/indi_allsky/flask/templates/config.html
+++ b/indi_allsky/flask/templates/config.html
@@ -1756,7 +1756,7 @@ var camera_id = {{ camera_id }};
                 Methods operate at similar speed. Except for Wavelet Filter- slowest, but best quality.</div>
             <ul>
             <li>Gaussian Blur &mdash; Targets noise with gentle, uniform smoothing.</li>
-            <li>Median Blur &mdash; Best for removing salt-and-pepper noise. Replaces outlier pixels with median of neighbors.</li>
+            <li>Median Blur &mdash; Best for removing salt-and-pepper noise. Outlier pixels replaced by median of neighbors.</li>
             <li>Bilateral Filter &mdash; Edge-aware filter. Smooths sky noise while preserving star edges.</li>
             <li>Wavelet Filter &mdash; Removes noise at all scales. Best for read noise and faint sky noise. Slowest, best quality.</li>
             </ul>
@@ -1782,7 +1782,7 @@ var camera_id = {{ camera_id }};
             {{ form_config.BILATERAL_SIGMA_COLOR(class='form-control bg-secondary') }}
             <div id="BILATERAL_SIGMA_COLOR-error" class="invalid-feedback text-danger" style="display: none;"></div>
         </div>
-        <div class="col-sm-8">Brightness difference to blend across (1-50). Low (10-20) preserves star edges tightly. High (40+) smooths aggressively. Default: 10</div>
+        <div class="col-sm-8">Brightness threshold for blending (1-50). Low (10-20) preserves star edges tightly. High (40+) smooths aggressively. Default: 10</div>
     </div>
 
     <div class="form-group row">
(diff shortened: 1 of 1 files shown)


ikeysolomon:
Forgot to mention that because of the time gating on the star mask function, the times quoted are much quicker during the day. Due to the star mask being bypassed. The processing times listed are for at night only.

aaronwmorris:
Sounds good.  I have some personal stuff going on, so it will be a few days before I take a look at this.  It looks very good though.

ikeysolomon:
Thats absolutely fine, take your time. The new directory is just scaffolding to support the PR it can be deleted for sure.

Interestingly, an observation is that the inherent star detection algorithm picks up more stars in my denoised images than not denoised. So id say the SNR of feint stars is being improved.

ikeysolomon:
image
**Figure 1:** Denoised 

image

**Figure 2:** Not denoised

Both images are 100% zoom-in of the same patch of day-time sky, denoised = using default bilateral filter settings

ikeysolomon:
image
**Figure 1:** Not denoised

image
Figure 2: Denoised

Both images are 100% zoom-in of the same patch of night-time sky, denoised = using default Wavelet filter settings.

PR #2793, opened 2026-03-17
BEHAVIOR CHANGE: add startrail timelapse scaling config

aaronwmorris:
This change adds a new, separate ffmpeg scaling option for star trail timelapses


aaronwmorris:
You should leave the night setting alone since you are already scaling the wrapped keogram images by 50%.  You only need to adjust the VFSCALE option for star trails.  I would suggest 1440px.

PR #2873, opened 2026-04-17
add scale limit for image registration

aaronwmorris:


illigtr:
Thanks!. Fast (as usual). Will update both of my indi-allsky cameras. Tonight is forecast as clear sky and new moon.

aaronwmorris:
I will have to correct this change.  I think the current state will work well enough to prevent images from becoming too small, but not from growing larger.

PR #2880, opened 2026-04-23
Fix startrails moonmode settings, fixes #2832

### LINKED ISSUE #2832: Star Trails Exclude Moon Mode - when inactive, star trails generation still relies on moon altitude and phase.
In order to produce star trails during moon mode, one must first activate "Exclude Moon Mode", then set the altitude and phase value to "disabled" as in image below. Save it, then go back and disable "Star Trails Exclude Moon Mode".

Image

tswfi:
The toggle is inverted and the fields should actually be disabled when the toggle is off.

Added a "inverted" parameter for the fields and made the texts a bit clearer.


tswfi:
doh, I wasn't on latest main. sec. I will double check :)

tswfi:
so this commit https://github.com/aaronwmorris/indi-allsky/commit/5c2a74168253c4bb09fd9a075c0de47b08c971e3#diff-a03dc2ef5764296135515113bda2d1cd2b345b3c240b3984722164c33e51a116 changed the text but didn't change the logic at all? 
I left my text as I think it is more descriptive but please let me know if you like that commit version better and I can change it.

PR #2886, opened 2026-04-24
Feature/fits selector

jerryfmedeiros:
The Problem

Currently, the Image Processing page always loads the most recent FITS image with no way to select a different one. This makes it difficult to test processing settings against a specific frame.

Changes Made
This PR introduces a FITS image selector to allow tuning against specific historical frames.

Queries the 200 most recent FITS images for the current camera.
Passes these images to the template as fits_image_list.
Adds a dropdown beside the hidden FITS_ID field. Populates the dropdown from fits_image_list, labeling each image by its capture timestamp. Selecting a different image immediately triggers a reprocess. Edge case handled: When no FITS images exist for the camera, the dropdown displays "No FITS images available".


jerryfmedeiros:
The FITS Viewer already has a Process link that passes the image ID directly to Image Processing . This PR adds a selector that duplicates that workflow

PR #2887, opened 2026-04-24
Feature/sep star detection

jerryfmedeiros:
**The Problem**
The existing star detection correlates each frame against a synthetic 15×15 blurred circle. This approach has some limitations on real allsky images: it assumes stars have a fixed shape that rarely matches a wide-angle lens, it lacks a sky background model (making it vulnerable to light pollution, clouds, and noise), and the threshold is a raw correlation value (0–1) with no physical meaning. The reported star count is often unreliable and difficult to tune.

**Changes Made**

- This PR adds SEP (Source Extractor Python) as an optional star detection method. SEP models and subtracts the varying sky background, then extracts sources based on a clear, physical sigma threshold (e.g., 5× the local noise).
- Implements the SEP detection algorithm as a drop-in option in the processing pipeline.
- Sizes the drawn marker circles proportionally based on the actual measured size of each star, so brighter/larger stars get larger markers.
- Updates the Settings UI with a new "Method" dropdown and dynamic threshold fields (enabling/disabling the appropriate threshold input based on the chosen method).
- Adds fallback handling for extraction errors and ensures the system handles mask conventions smoothly.
- Uses lazy loading for the new library, meaning the sep package is only imported and required if the user actively selects the SEP method.
- Backward compatibility: Preserves the existing template method as the default, ensuring no changes or disruptions for current users unless they explicitly switch.

Dependency Note: Requires pip install sep>=1.4.1 to use the new method.


aaronwmorris:
I have not forgotten about this.  This PR was submitted during a time when I was dealing with some personal issues that are mostly resolved now.

Unfortunately, with the new UI work, this will no longer cleanly merge.  I am still interested in integrating this, but it would need to be reworked with the new interface.

jerryfmedeiros:
Give me a couple of weeks, I can familiarize myself with the new UI and port it.

TN-1:
Let me know if you have any questions with any of it :)

jerryfmedeiros:
This was much easier than I anticipated. The config page being split into per tab partials made it much simpler than editing the old beast, and the card layout is nicer to work with. Really like the new interface, looking forward to running it on my own instance.


Settings now live in templates/config/processing.html instead of the old config.html, inside the existing Star & Meteor Detection card. Three fields: Method, SEP Sigma Threshold, SEP Max Star Radius. Only the threshold for the selected method stays editable, the other greys out.

The detection code itself is unchanged from the original PR.

Sep turns out to already be installed on every indi-allsky system. It is a dependency of astroalign, which is already required. So adding it explicitly to requirements costs nothing at install time. It is still imported lazily, so platforms where it cannot build (armv6l, debian10) keep working on the template method.


### Template Method (Default - SEP settings grayed out) 
Screenshot 2026-08-21 at 12 51 24 PM

### SEP Method 
Screenshot 2026-08-21 at 12 51 36 PM

### Comparison
PR_star_detection_comparison


Same patch of sky, same detection mask, thresholds picked so both methods return almost the same count (230 vs 229). Green means there is a real source at that position (a pixel within 3px that is more than 4 sigma above the local sky). Red means empty sky.

Template: 230 detections, 22 false positives, 90.4% precision.
SEP: 229 detections, 2 false positives, 99.1% precision.

@aaronwmorris Let me know what you think and also if you want me to work on public documentation. Not sure what's the best way to contribute to the Wiki.

PR #2888, opened 2026-04-24
Feature/detection tuning tab

jerryfmedeiros:
**The Problem**

Currently, there is no easy way to tune detection thresholds interactively. The only workflow is to change the threshold in Settings, save, wait for the next capture, and check the logged star count. 

**Changes Made**
This PR adds a Detection tuning tab to the Image Processing page, providing immediate visual feedback for threshold tunning.

- Adds a "Detection" tab (styled consistently as a tuning tool) to the Image Processing interface.
- Seeds the tuning fields with current configuration values so the tab opens with meaningful defaults.
- Introduces a "Run Detection" toggle and result badges for star and meteor counts.
- When detection is enabled, it runs star and meteor detection on the processed preview image and draws the results directly on the image in-place.
- Dynamically updates the UI after each request to display the live detection counts and automatically shows or hides the relevant threshold field based on the currently active detection method (Template vs. SEP).

 Dependency Note: This PR depends on PR #2887 (feature/sep-star-detection) as the new tab supports tuning for both detection methods.


aaronwmorris:
I am still interested in this, as well.

jerryfmedeiros:
I can port it as well.

jerryfmedeiros:
Ported to the new UI as well. Stacked on #2888,  so please merge that one first. The diff here will shrink to just the tuning card once it lands.

### Where I put it


I made it a card inside the **Image** tab of the Image Processing page rather than a fifth top level tab. The detection run reads DETECT_MASK, which already lives on that tab, so keeping them together avoids splitting related controls. And the tab strip you designed reads to me more as categories of settings, while this is more of a tool.

### What it does

Toggle Run Detection, hit Process, and it runs star and meteor detection on the preview and draws the results with live counts. Nothing is saved, it all runs against a copy of the config.

I added a Method dropdown to the card too, which was not in the original PR. Without it you have to go to Settings, change the method, save, and come back just to compare the two. Now you can A/B them in a couple of clicks.

Two bugs I hit while testing

- Meteor counts were being polluted by star detection. The original code ran star detection first, which draws circles on the image, then ran line detection on that same image, so Hough was counting the star markers as lines. On a real frame that was 543 meteors at sigma 5 versus 0 at sigma 15, all from the markers. image.py runs lines first then stars, so I matched that order and the count is stable now.

- The other one is not related to this PR but it blocks it. detectLines._drawLines assumes HoughLinesP returns shape (N, 1, 4), which changed to (N, 4) in OpenCV 5, so it raises a TypeError. Since requirements_latest.txt has opencv-python-headless >= 4.10.0.84 with no upper bound, new installs get OpenCV 5 and hit this. Anyone running meteor detection with DETECT_DRAW on is affected in the normal capture pipeline, not just here. It is a one line fix using numpy.atleast_2d. Happy to pull it into its own PR if you would rather review it separately.


Screenshot 2026-08-21 at 1 09 44 PM
Screenshot 2026-08-21 at 1 10 02 PM
Screenshot 2026-08-21 at 1 10 08 PM


@aaronwmorris  let me know what you think. Happy to move it around if you prefer it on its own tab or somewhere else. 
Thanks!

aaronwmorris:
@TN-1 This is an example of the `dist.css` conflict.

jerryfmedeiros:
@aaronwmorris @TN-1  I regenerated it for now, but it will keep happening. 

One suggestion to stop it is to add a `.gitattributes`:

`indi_allsky/flask/static/css/dist.css merge=ours`

 and one more bit to `prepare` in package.json:

`"prepare": "git config core.hooksPath githooks && git config merge.ours.driver true || true"`

I tried this on a test repo: Without it the rebase stops on the conflict. With it the rebase goes through and the file keeps upstream's version, which the pre-commit hook rebuilds on the next commit anyway.

TN-1:
You just need to delete the dist.css and rebuild it on a rebase. That fixes the conflicts:

``` bash
# During rebase/merge conflict on dist.css:
rm indi_allsky/flask/static/css/dist.css
npm run build
git add indi_allsky/flask/static/css/dist.css
git rebase --continue
```

Just the nature of the big css file, so having to rebase the latest changes from main, then rebuilding the dist.css is the best way forwards. It is complied based on whats in the branch at the time, so when Upstream Main brings in changes, dist.css gets out of sync.

We could try doing the .gitattribute, but then we are just masking the error (IMO) and if someone forces a commit without rebuilding dist.css, then we end up in a bad place again with dist.css out of sync with the rest of the code base. I think there is a better solution possible.

Let me look into it some more, and I will see what I can come up with for a better long term fix to make life easier.

jerryfmedeiros:
> You just need to delete the dist.css and rebuild it on a rebase. That fixes the conflicts:
> 
> ```shell
> # During rebase/merge conflict on dist.css:
> rm indi_allsky/flask/static/css/dist.css
> npm run build
> git add indi_allsky/flask/static/css/dist.css
> git rebase --continue
> ```
> 
> Just the nature of the big css file, so having to rebase the latest changes from main, then rebuilding the dist.css is the best way forwards. It is complied based on whats in the branch at the time, so when Upstream Main brings in changes, dist.css gets out of sync.
> 
> We could try doing the .gitattribute, but then we are just masking the error (IMO) and if someone forces a commit without rebuilding dist.css, then we end up in a bad place again with dist.css out of sync with the rest of the code base. I think there is a better solution possible.
> 
> Let me look into it some more, and I will see what I can come up with for a better long term fix to make life easier.

Fair point.  If the worry is dist.css drifting out of sync, that might be worth catching in CI rather than relying on everyone remembering. Something like running npm run build and failing if git diff --exit-code on dist.css is dirty. Then a stale build gets caught on the PR instead of landing.

TN-1:
Yeah thats a thought I have had, but then if we move the build into CI then anyone testing a branch will need to build it themselves. Pros and cons everywhere, lol.

Leave it with me, I will see what I can come up with.

TN-1:
Have a look at #3085

PR #2902, opened 2026-04-29
ensure gamma correction is not 0

aaronwmorris:


callimero:
Oh so, "User to stupid" (overwhelmed by the options). Thanks!

aaronwmorris:
Definitely not user error.  I do try to put extra effort into ensuring the settings cannot break the algorithms.

PR #2903, opened 2026-04-29
White Balance Early in Processing Pipeline

EthanChappel:
White balance currently happens after stretching, but should be done while the image is linear to keep accurate colors.

Used ColorCalibration in PixInsight to find WB factors with a FITS image using the window trim as white reference and got a pleasing result. Applying them in indi-allsky gives me an image with a pink sky.
ccd1_20260429_145138-480

It never looks quite right, even if I try manually balancing the above image. Some part always has a strange tint.<br>

Moving WB earlier in the processing pipeline gives me the expected result straight from indi-allsky:
ccd1_20260429_144648-480

This is with the same settings as the first image.

Anyone who adjusted their WB factors would see their image change after updating. Whether that needs to be addressed could be worth discussing, but overall I think this improves white balancing.

### DIFF
--- a/indi_allsky/image.py
+++ b/indi_allsky/image.py
@@ -577,6 +577,12 @@ class ImageWorker(Process):
         # adu value may be updated below
 
 
+        # white balance
+        self.image_processor.white_balance_mtf()
+        self.image_processor.white_balance_manual_bgr()
+        self.image_processor.white_balance_auto_bgr()
+
+
         self.image_processor.denoise()
 
         self.image_processor.stretch()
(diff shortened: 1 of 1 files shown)


aaronwmorris:
I will hold on to this for a while.  While not exactly a breaking change, it would alter the current processing order and change the established defaults other people currently have.

aaronwmorris:
To be honest, there really needs to be a way to reorder the processing pipeline.

EthanChappel:
Sounds like a better solution in the long term.

Issue #2905, opened 2026-05-07
Chart/overlay labels are hardcoded "Topic #" using MqttBrokerSensor


jerryfmedeiros:
Every chart/overlay slot for an MqttBrokerSensor renders as "Topic 1" to "Topic 10" in `mqttBrokerSensor.py:19-30`.
                                                                      
  The configured topic (`A_PIN_1`...) can be a human-readable label,  but there's no way to surface it, we need to users keep a side cheatsheet to know which "Topic N" is which.           

<img width="395" height="691" alt="Image" src="https://github.com/user-attachments/assets/5833f3d0-c5bd-4a0b-9de2-287dce760865" />

aaronwmorris:
Because of the way the labels are read for sensors, the only way to update them for the MQTT sensor is to edit the code.

https://github.com/aaronwmorris/indi-allsky/blob/main/indi_allsky/devices/sensors/mqttBrokerSensor.py#L20

jerryfmedeiros:
Makes sense. Submitted a PR (#2906 ) with a possible solution.  I am using that on my instance where I have dozens of external sensors for weather and micro-climate research. 

<img width="422" height="560" alt="Image" src="https://github.com/user-attachments/assets/ad7eea2d-847a-469c-b0ec-c9dc3ebab7e7" />


I am relying on the  ESPHome / Home Assistant convention: "<prefix>/sensor/<name>/state" that might not be true for everyone. We could include it in the documentation. It wouldn't crash or anything, just fall back to the original labels.  

Let me know what you think.

jerryfmedeiros:
Hey @aaronwmorris since the PR #2906  was merged, I realized the Wiki documentation for the MQTT sensor is slightly out of date now that the dynamic labels are working. Since I can't open a PR for the wiki directly, I drafted an updated section for that page. Feel free to copy paste this in whenever you have a chance:

MQTT Broker Sensor
A sensor value from an MQTT Broker may be queried. The Pin field is used for the topic subscription. Up to 10 topics may be subscribed to by providing a comma-separated list (e.g., topic1, base/topic2).

The payload for the topics must be a value that can be converted to a float.

[...]

PR #2906, opened 2026-05-07
MqttBrokerSensor:  grab chart/overlay labels from configured topic names

jerryfmedeiros:
Replaces the static `Topic 1..10` labels with names derived from the user's configured topics.

 ## Changes                                                                    

  - `mqttBrokerSensor.py`: new optional classmethod `get_dynamic_labels(pin_1_name)` — formats the comma-separated topic list,  drops trailing `/state`, pads to 10 with `Topic N`.                           

  - `capture.py`: `update_sensor_slot_labels()` reads `*_PIN_1` for slots A–F and uses `get_dynamic_labels` when the driver class implements it. Falls back  to `METADATA['labels']` otherwise.                                            
                                                  

### Backward compatatibility                                                            

  No new config keys, no schema changes. Every other driver lacks `get_dynamic_labels` → existing path unchanged (verified against BME280, INA228, MPU6050, simulator).

Solves: #2905


aaronwmorris:
This makes sense.  I will take a closer look in a bit.

PR #2993, opened 2026-07-23
feat: add authenticated browser-based terminal shell window

### LINKED ISSUE #2929: Feature Request: Add "Open Shell Window" option to Tools menu?
I frequently find a need to get to a shell window for various things (like running htop, etc.) and it would be nice (and easy to add?) to have an option to open a shell window from say the Tools menu? This would really be helpful when using a browser from a remote machine, not the allsky server itself (obviously, it's pretty easy and fast to open a shell window directly from the allsky machine)

Thank you

TN-1:
- Add '/shell' UI view and '/shell/ws' websocket route using simple-websocket and pty
- Include local xterm.js assets and fit addon to render terminal in browser
- Restrict route access and sidebar visibility to authenticated users
- Update Nginx and Apache server templates to support websocket upgrading
- Add simple-websocket to Python requirements files
- Remove PrivateTmp=true from Gunicorn service to allow setuid/sudo usage inside shell
- Add caution warning banner upon starting interactive shell session

image

Closes #2929


indi_allsky/flask/templates/shell.html
```diff
@@ -0,0 +1,168 @@
...
+    } else if (!wsUrl.startsWith('ws://') && !wsUrl.startsWith('wss://')) {
+        // Relative path fallback
+        const loc = window.location;
+        const wsProtocol = loc.protocol === 'https:' ? 'wss:' : 'ws:';
+        wsUrl = wsProtocol + '//' + loc.host + (wsUrl.startsWith('/') ? '' : '/') + wsUrl;
+    }
+    const socket = new WebSocket(wsUrl);
+
+    socket.onopen = function() {
+        term.write('\x1b[1;31mWARNING: Interactive shell session. Commands are executed directly on the host system.\x1b[0m\r\n\r\n');
```
TN-1, line 107: Do we like this warning or want to change it?

service/gunicorn-indi-allsky.service
```diff
@@ -12,7 +12,6 @@ ExecStart=%ALLSKY_DIRECTORY%/virtualenv/indi-allsky/bin/gunicorn --config %ALLSK
 ExecReload=/bin/kill -s HUP $MAINPID
 KillMode=mixed
 TimeoutStopSec=5
-PrivateTmp=true
```
TN-1, line 15: @aaronwmorris Was PrivateTmp=true needed for anything in particular? If we keep it in, then sudo doesnt work in the shell (Among other bugs that may exist I havent found yet). Removing it fixes the problem and allows priv escalation in the shell.

aaronwmorris: I do not believe I had a specific requirement for it.  It was just a security enhancement, but if there is a reason to disable, that is fine with me.

indi_allsky/flask/templates/shell.html
aaronwmorris: That works for me

aaronwmorris:
Is this code basically ready to merge?  I have a few minor alterations to make, but I would like to begin integration.

TN-1:
Yup, go for it

aaronwmorris:
The only major change I made was changing the route from `/shell/ws` to `/ws/shell`.  I just did this in case there are any future websocket views, they can all be put under the `/ws` path.

TN-1:
Yeah thats a good call

PR #2997, opened 2026-07-24
Feature - Support for esp32 round disc allsky viewers

### LINKED ISSUE #2693: Feature - Support for esp32 round disc allsky viewers
I recently acquired and built an allsky viewer disc and ran into some issues with indi-allsky.

if I was running regular allsky, I could use this module which would do what I need: https://github.com/AllskyTeam/allsky-modules/blob/master/allsky_esp32round/allsky_esp32round.py

Essentially it needs to:
1)   Automatically resize the latest image to 720x720. or 800x800 depending on the model of display disc. It only supports JPEG format, no webp or anything.
2) Offer an actual static image link to download the image.  The esp32 firmware being used doesn't support HTTP 302, so the URL provided to it has to resolve immediately to the image in the correct size/format.
3) The image can't require authentication to load.

The project in question is located here, which includes what sizes are what resolution, etc. https://github.com/chvvkumar/ESP32-P4-Allsky-Display/tree/main

harshareddy832:
Fixes #2693

Changes:
- Addressed the issue described in #2693
- Followed existing code conventions


aaronwmorris:
I do not mind merging this PR, but was there something wrong with the existing implementation in indi-allsky for generating images for the circular displays?

harshareddy832:
Thanks for the review, Aaron!

The existing implementation wasn't broken per se, but the issue was that it only served the latest captured image — which requires the full indi-allsky pipeline to be running and producing images. For circular displays, I wanted to support a dedicated pre-processed image (`circular_display.<ext>`) that can be generated separately and served directly, without depending on the latest image entry being available.

This is useful when you want to run a lightweight setup where the circular display image is updated via its own process, rather than always pulling from the main image pipeline. The fallback still works if the circular display config isn't enabled.

Happy to revert if you'd prefer to keep things in the existing flow — just thought this gave more flexibility.

aaronwmorris:
That makes sense.

PR #3010, opened 2026-08-09
New Web UI Interface and Theme

aaronwmorris:
Thank you, @TN-1


TN-1:
Nice catch with [f85275d](https://github.com/aaronwmorris/indi-allsky/pull/3010/commits/f85275dc8070845be8d7c1219c883c21deeb9d19) :)

aaronwmorris:
I just happened to notice the 404 errors in the web server log.  I am not sure that it really matters, the site was working fine with the wrong path.  I guess this is a fix, but there is an edge case where the `/indi-allsky/` context root can be changed.   I guess in that case, someone might just get some 404 errors in their log.

TN-1:
> I just happened to notice the 404 errors in the web server log.  I am not sure that it really matters, the site was working fine with the wrong path.  I guess this is a fix, but there is an edge case where the `/indi-allsky/` context root can be changed.   I guess in that case, someone might just get some 404 errors in their log.

Problem for a future day 🤣

PR #3019, opened 2026-08-10
Add ASI676MC purple-frame detection, repair, and calibration

321MathiasS123:
Implements the optional ASI676MC purple-frame handling developed in [Discussion #2987](https://github.com/aaronwmorris/indi-allsky/discussions/2987). See that discussion for the problem background, repair method, examples, and screenshots. I've tried to make this as robust as possible and this has made the PR quite substantial.  This has been tested as much as is possible with a single installation and one affected camera. In theory one should be able to point this at an existing indi-allsky database containing fits from an ASI676MC and press "Go", then either apply the derived values directly or apply the corrected detection thresholds and run again. Manual upload assumes at least some file selection by the user but will also discover bad frames automatically if the fits are not marked by the detection.

### Summary

- Adds opt-in detection with Repair and Exclude Only modes.
- Strictly limits processing to compatible ASI676MC RAW16/RGGB captures.
- Adds a calibration tool using either uploaded FITS or the complete retained-FITS database.
- Supports legacy and unmarked FITS, automatic population discovery, cancellation, reporting, and direct application of calibrated settings.
- Prevents excluded or validation-failed frames from affecting stacking and exposure control.
- Includes user documentation and focused automated tests.

The feature is disabled by default and does not change existing behavior for other cameras.

### Validation
- 123 focused repair and calibration tests passed.
- Python compilation and diff validation passed.
- Automatic discovery was exercised with 10,000-record metadata and legacy-FITS scenarios.
- Manual and database calibration paths were checked against the same real FITS collection, with the same compatible and incompatible files identified.

Additional testing with other ASI676MC cameras and datasets is welcome.


321MathiasS123:
@aaronwmorris 
I think it's ready for review and merging.

321MathiasS123:
Needs one final pass for the documentation.. then it's ready

321MathiasS123:
And done.

PR #3035, opened 2026-08-12
VirtualSky: astrometric auto-calibration (Solve button)

jaxzin:
## Why

Aligning the VirtualSky overlay by hand means iterating six interacting numbers (azimuth angle, lat/long offsets, image circle diameter, X/Y offsets) against the sky image — adjust, reload, compare, repeat. It took me over an hour of tinkering to get an alignment I was happy with, and any bump to the camera means doing it again.

This PR adds a **Solve** button to the VirtualSky page that fits all six values from the image being displayed, in a second or two. The operator reviews the result (the overlay re-renders immediately) and only then clicks **Save**. On my camera, the values I had hand-tuned turned out to be ~4° off in azimuth and ~90 px off-center — the solved overlay is visibly better than what I'd settled for.

## How it works (short version)

Detects stars in the displayed frame (OpenCV), matches them against VirtualSky's own star catalog, and least-squares fits **VirtualSky's exact rendering model** — equisolid fisheye, same GMST convention, no precession/refraction — so the solution aligns with what the renderer actually draws rather than textbook astrometry. Runs entirely locally; no astrometry.net, no network.

## Measured runtime (real 3552×3552 frame, 8.3 MP camera)

| Hardware | Typical solve | Worst case (azimuth 180° off AND diameter 16% off) |
|---|---|---|
| Raspberry Pi 4B (my production allsky host, under normal capture load) | 0.7 s | 2.1 s |
| Raspberry Pi 5 | 0.3 s | 0.8 s |
| x86 dev machine | 0.4 s | 1.8 s |

Memory use is bounded for 512 MB hosts by downscaling detection above 16 MP.

## Scope and safety

- **No new dependencies** — numpy/scipy/opencv/astropy only, already required on every platform including armv6l.
- Solve requires a login; Save is only available to admins and writes `LENS_AZIMUTH` plus the five `VIRTUALSKY` geometry keys.
- Only one solve runs at a time — a second request while one is in progress gets an immediate HTTP 429 "busy, try again" rather than piling up work (assumes the shipped one-worker gunicorn config).
- Solves the *displayed* image, including historical frames opened via the Gallery — so calibration works in daylight from last night's data.
- Poor frames (clouds, moonlight, twilight) are refused with a specific reason, never a confident wrong answer; pure-noise inputs are regression-tested to fail. A mirrored image is detected and reported with the exact config flip to make.
- Recovers from grossly wrong starting values (any azimuth, diameter ±40%); only the X/Y offsets need to be within ~100 px.
- 178 unit tests (`tests/test_lens_solver_*`); the star catalog is extracted from the already-vendored virtualsky.js + stars.json by a committed script (`testing/generate_lens_solver_catalog.py`).

Happy to adjust anything — naming, placement, gating — to fit the project's direction, and to write the wiki section for it if that's useful.


TN-1:
This is very cool, nice work

aaronwmorris:
I have to agree.  This is basically magic "fix it" button!

Issue #3040, opened 2026-08-13
Docs: Star history chart is broken


TN-1:
Nice timing, right after we added it to the README github broke the functionality it depends on. Needs you to generate an API key to continue working Aaron. 

See: https://www.star-history.com/blog/github-stargazer-api-restriction/

PR #3041, opened 2026-08-13
Fix broken star history chart in README

Dessalines39394:
The star history chart in the README no longer renders because the upstream star-history service relies on the GitHub stargazer API, which is now restricted and requires an API token. Switch the chart to a working mirror of the same service so it displays correctly without any token. This keeps the chart's visual output identical while restoring the broken functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -768,11 +768,11 @@ npm run build
 
 ## Star History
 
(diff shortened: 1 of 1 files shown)


TN-1:
Closes #3040. Nice solution

aaronwmorris:
Thank you!

Issue #3030, opened 2026-08-11
Various anomilies with new UI - August 2026 version


illigtr:
First off, congratulations to you and your contributors on this major upgrade. Probably far more work than anyone knows.

I have tried the new UI on my smartphone (Galaxy S21FE), my tablet (Galaxy Tab 6 Lite) and my desktop computer (Win 11).

### First impressions
The UI on my smartphone feels like a major regression, it is far more difficult to navigate and get anything done at all. The menu items are collapsed to only one item showing at a time (Latest, or Loop...). Attempting to use or change config is just too difficult.

On my tablet, the experience is far better, probably  due to enough screen room that the UI becomes usable.

On my desktop, the UI is smooth enough. However, using it with Firefox causes all sorts of problems, mostly buttons won't click. Switching to Edge in Win 11 allowed the buttons to function normally.

### General and Quirks
Although I have sign in ONLY for config changes, the new GU forces a sign in for read only now.

In config, there used to be an option to "Save and Reload". Doesn't seem to exist, thus requiring an extra step to go to System / System Info/ Services and reload from there.

aaronwmorris:
A lot of the issues with clicking not working can be resolved by flushing your browsers cache.  There is some mechanism that is conflicting between the old Bootstrap UI and the new Daisy UI.

The "Save and Reload" feature has been re-added with #3028

illigtr:
Thanks! I'll try those suggestions.

TN-1:
> On my desktop, the UI is smooth enough. However, using it with Firefox causes all sorts of problems, mostly buttons won't click. Switching to Edge in Win 11 allowed the buttons to function normally.

This sounds like the cache bug, Im assuming you dont use edge normally? Flush the cache as Aaron said and that will solve most things.

> The UI on my smartphone feels like a major regression, it is far more difficult to navigate and get anything done at all. The menu items are collapsed to only one item showing at a time (Latest, or Loop...). Attempting to use or change config is just too difficult.

The behaviour of the sidebar on mobile and desktop at the same. On mobile, the sidebar is responsive and collapses to allow more screen space for content. Unsure how you mean config is too difficult, I find the new mobile UI to be alot more usable than the old one. Try clearing the cache on your mobile browser too in case you are experiencing issues with the cache bug. (Though I can appreciate some people would prefer the older UI of zooming and panning across the desktop UI on mobile vs having something designed specifically for a smaller screen thus needing more scrolling.)

[...]

PR #3042, opened 2026-08-13
Fix/gallery scroll and sticky UI

TN-1:
Fixes a couple of issues raise in #3030

Restores previous behaviour in gallery with regards to window scrolling to the top when the hour is changed by keys
Stickys the date picker bars on videoviewer and gallery, and the save bar in config
Fixes virtualsky viewport sizing (now fits to screen, as it was before)


indi_allsky/flask/templates/base.html
```diff
@@ -113,6 +113,14 @@
                 display: block;
             }
         }
+
+        @media (min-width: 768px) {
```
TN-1, line 117: Might be worth considering whether we want this behaviour, or if we want to sticky them on mobile too. I am unsure what the best option is and am interested to hear thoughts on it.

PR #3048, opened 2026-08-14
Make panorama timelapse and loop respect image exclusion

321MathiasS123:
Even before working on the purple frame correction I noticed that the existing `image.exclude` flag is not carried over to the panorama loop and automatic panorama timelapses. Panoramas use a separate `panoramaimage.exclude` flag, but that can't be set in any way through the UI right now.  It's highly unlikely that a user manually excludes an image from the normal timelapse but wants that in the panorama timelapse. 

Since there is no foreign key linking a panorama back to the source image and I don't know whether `panoramaimage.exclude` is intended for something, this little implementation checks the source image based on `camera_id` and `createDate`. If either one or both exclude flags are set to true, the image will be excluded from the panorama loop and timelapse.

This also makes one paragraph in the purple frame handling doc obsolete, so that's removed, too.


aaronwmorris:
This seems like a pretty complex integration.  Couldn't we just set the panorama exclude flag at the same time as the image exclude?  The timelapse generation and Loop still obey the exclude flag.

321MathiasS123:
OK, I'll change that over. I didn't know what to do with the existing exclude flag specific to panoramas. Let's see how that goes, the panorama rows don't directly refer back to the source image once they are created.

321MathiasS123:
- New panorama rows inherit the image exclusion value. When the panorama is generated the exclusion value at capture time is used. This allows the purple-frame-handling to exclude those from the start.
- Manual include/exclude changes update the matching panorama row.
- Panorama loop and timelapse now use only `panoramaimage.exclude.`
- Removed the cross-table query helper
- Existing panorama rows are not automatically backfilled, although toggling their corresponding image will synchronize them.

PR #3064, opened 2026-08-18
Debian Packaging and CI/CD Action + Dependabot

TN-1:
This has been a slog for the last week, lots of tweaking and tracking down packaging bugs, but I feel this could be a huge improvement over the current `setup.sh`

Introduces deb packaging and an automated multi-arch/distribution CI/CD pipeline. This eliminates the need for slow compilation scripts on target hardware (such as Raspberry Pi) and replaces `setup.sh` with a modern system installation, allowing for easier upgrades and faster installs 

### Debian Packaging (`debian/`)
* **Dedicated Service User**: Runs under isolated system user/group `indi-allsky:indi-allsky` with proper hardware groups (`video`, `dialout`, `plugdev`, `gpio`).
* **Offline Python Wheel Cache**: Bundles pre-compiled wheels for all 100+ dependencies (including `astropy`, `opencv`, `scipy`, `cryptography`, and SWIG `PyIndi` bindings) into `/usr/share/indi-allsky/wheels/`.
* **Interactive Debconf Wizard**:
  * Step-by-step configuration for camera interface, CCD driver, ports, observatory coordinates, and admin user credentials.
  * Interactive password confirmation and length ($\ge 8$ chars) validation.
  * Preserves existing `/etc/indi-allsky/flask.json` settings across package upgrades.
* **Automated Web Server & Proxy Integration**: Automatically installs and provisions Apache2 or Nginx with self-signed SSL SAN certificates and WebSocket reverse-proxy routing.
* **Systemd Service Management**: Automatically provisions, enables, and manages `indi-allsky.service`, `indiserver.service`, and socket-activated `gunicorn-indi-allsky.service`.

### 2. Multi-Architecture CI/CD Pipeline (`.github/workflows/package-deb.yml`)
* **Multi-Distro Matrix**: Automatically compiles and packages for:
  * **Debian 12 (Bookworm)**
  * **Ubuntu 24.04 (Noble)**
  * **Debian 13 (Trixie)**
* **Multi-Arch**: Builds natively for both `arm64` (Raspberry Pi 4/5, aarch64) and `amd64` (x86_64).
* **Automated QA  Tests**: Every built `.deb` is automatically verified in a fresh container (validates package installation, virtualenv C-extension imports, database schema bootstrap, admin user creation, and clean purge).
* **Automated Releases**:
  * **Nightly**: Builds run automatically every night at 02:00 UTC and attach to pre-releases with commit changelogs.
  * **Stable**: Version-tagged pushes (`indi_v*`) publish stable `.deb` releases with curated release notes.

### 3. Automated Dependency Maintenance (`.github/dependabot.yml`)
Added automated monthly dependency updates with grouped pull requests to keep dependencies fresh without creating PR noise:
  * **Python (pip)**: Monitored in `/requirements`
  * **Frontend (npm)**: Tailwind CSS, DaisyUI, Cally, and Iconify plugins
  * **CI/CD**: GitHub Actions versions
  * **Containers**: Docker base images & Docker Compose definitions

### 4. Version Management & Release Steps
Added `misc/bump_version.py` as a single source-of-truth utility that synchronizes `indi_allsky/version.py`, `package.json`, `debian/changelog`, and template cache-busters.

**How to Cut a New Release:**
1. Bump version across all files:
   ```bash
   python3 misc/bump_version.py 2026.08.2
   ```
2. Commit and tag:
   ```bash
   git commit -am "chore(release): bump version to 2026.08.2"
   git tag indi_v2026.08.2
   ```
3. Push to GitHub:
   ```bash
   git push origin main --tags
   ```
4. GitHub Actions will automatically build all distribution packages, run QA tests, and publish the `.deb` assets to the GitHub Release page.

---

#### 1. Enable GitHub Actions Workflow Permissions
To allow GitHub Actions to publish GitHub Releases, attach `.deb` assets, and deploy to GitHub Pages:
1. Go to repository **Settings** > **Actions** > **General**.
2. Scroll down to **Workflow permissions**.
3. Select **"Read and write permissions"** (allows `GITHUB_TOKEN` to create releases and deploy artifacts).
4. Check **"Allow GitHub Actions to create and approve pull requests"**.
5. Click **Save**.

#### 2. Enable GitHub Pages (for APT Repository Hosting)
To host the APT package repository and landing page via GitHub Pages:
1. Go to repository **Settings** > **Pages**.
2. Under **Build and deployment** > **Source**, select **"GitHub Actions"** (do *not* choose "Deploy from a branch").
3. *(Optional)* If using a custom domain (e.g., `apt.indi-allsky.org`), configure it under **Custom domain** and check **Enforce HTTPS**.

#### 3: Generate GPG Signing Key

The APT repository metadata (`Release` / `InRelease` file) is cryptographically signed using GPG so `apt` verifies package authenticity.

On your local machine, generate an RSA 4096-bit GPG key (with `%no-protection` so it can sign packages unattended in GitHub Actions CI/CD):

```bash
gpg --batch --gen-key <<EOF
%no-protection
Key-Type: RSA
Key-Length: 4096
Subkey-Type: RSA
Subkey-Length: 4096
Name-Real: INDI Allsky Repository
Name-Email: <your-email-here>
Expire-Date: 0
EOF
```

List your secret keys to find your Key ID:

```bash
gpg --list-secret-keys --keyid-format LONG
```
*(Look for the string after `rsa4096/`, e.g. `sec rsa4096/3AA5C34371567BD2`).*

#### 4: Configure GitHub Repository Secrets

Export your private key:

```bash
# Export using your Key ID or Email
gpg --armor --export-secret-keys <YOUR_KEY_ID_OR_EMAIL>
```

1. Navigate to your repository on GitHub: **Settings** -> **Secrets and variables** -> **Actions**.
2. Click **New repository secret**.
3. Add secret:
   - **Name**: `GPG_PRIVATE_KEY`
   - **Value**: Paste the exported GPG private key block (including `-----BEGIN PGP PRIVATE KEY BLOCK-----`).

---

### Tested:
* **Fresh VM Install (AMD64 / Debian 12)**: Verified complete installation from `.deb` packages.
* **Daemon & Service Stack**: Verified `indi-allsky.service`, `indiserver.service` (with `indi_simulator_ccd`), `gunicorn-indi-allsky.socket`, and `apache2.service` running cleanly.
* **Web UI & Reverse Proxy**: Verified login page and web dashboard over HTTPS with WebSocket streaming.
* **Automated CI QA**: 100% passing non-interactive tests in GitHub Actions containers for Debian 12, Debian 13, and Ubuntu 24.04.

### Testing Still Required (Seeking Testers):

- [x] **Physical `arm64` Hardware**: Installing and running on Raspberry Pi 4 / Raspberry Pi 5 with physical camera hardware (ZWO ASI, PlayerOne, SVBony, QHY, or libcamera).
- [x] **Migration from `setup.sh`**: Testing `.deb` installation on systems previously configured via `setup.sh` (verifying user-level systemd service migration and database preservation).
- [x] **In-Place Package Upgrade**: Installing a newer version `.deb` over an existing `.deb` installation to verify seamless database migrations and config retention.
- [x] **Nginx Web Server**: Testing installation on systems with Nginx instead of Apache2.
- [x] **Docker**: Verifies that the updated dockerfiles still work as expected

---

## Quick Install for Testers
https://apt.indi-allsky.org/


TN-1:
This also compiles and includes the various INDI dependencies that needed manual compilation on Pi's. ~~It currently pulls the latest version of the INDI repos at build time, but we can add some fields to define which version we want to build with~~ I have added pins for the INDI versions, and added a workflow to check weekly for new versions and open a PR

TN-1:
Another fun possibility this opens up is the creation of a small indi-allsky apt repo. Then users could just add the repo, do `sudo apt update`, `sudo apt install indi-allsky` and `sudo apt upgrade`

TN-1:
If you approve this Aaron, one thing that will need consideration is future support of setup.sh. IMO, the path that makes most sense would be to maintain both deb (preferred) and setup.sh (deprecation warning) until some change is made that needs setup.sh to be updated, at which point we remove it. Maintaining both install paths would be annoying and could lead to feature drift

TN-1:
### Debian Packaging Infrastructure (debian/)

- **control**: Defines package metadata, architecture support, maintainer details, and required system dependencies.
- **rules**: Makefile executed by debhelper to compile wheels, bundle offline assets, and install systemd units.
- **config**: Debconf script providing the multi-step interactive setup wizard and input validation.
- **templates**: Prompts, questions, descriptions, and error dialogs displayed by Debconf during installation.
- **postinst**: Post-installation script that provisions system users, sets up virtualenv, bootstraps SQLite, configures SSL, and restarts web servers.
- **prerm**: Pre-removal script that cleans up dynamically generated __pycache__ bytecode files before package deletion.
- **postrm**: Post-removal script that cleans up directories, services, and debconf database entries upon package purge.
- **format**: Specifies Debian source format (3.0 (native)).
- **copyright**: Machine-readable copyright and GPL-3.0 licensing specification.
- **README.Debian**: Debian package documentation explaining directories, architecture, and configuration files.
- **qa_test.sh**: Automated container test runner that verifies installation, venv C-extensions, SQLite auth, and package purge.

### CI/CD & Automation (.github/, misc/)

- **package-deb.yml**: GitHub Actions workflow compiling multi-distro/multi-arch .deb packages, running QA tests, and publishing releases.
- **dependabot.yml**: Dependabot configuration providing automated monthly dependency updates across Python, NPM, Docker, and GitHub Actions.
- **bump_version.py**: Single source-of-truth CLI utility synchronizing versions across Python, package.json, debian/changelog, and templates.

### Documentation (docs/)

- **debian_deployment_workflow.md**: Architectural documentation explaining the packaging pipeline, wheel caching, and driver consolidation.
- **setup_to_deb_migration.md**: Administrator migration guide for transitioning existing installations from setup.sh to native .deb packages.

TN-1:
Known issues:
~~When migrating from a setup.sh install to a deb install, admin account doesnt work. No investigation done yet~~
Time status not showing

TN-1:
After a sleep and with some fresh eyes, this PR has become quite expansive. Might be worth removing the docker changes for now and do them in a future PR since that itself is quite a large change

aaronwmorris:
I will review the changes.

I was reviewing the docker related changes before you removed them.  I would not have been able to accept that because it removed all of the hardware camera support.  One of the changes tried using the Debian libindi packages which are too old to properly function.

TN-1:
Yeah the docker changes were going to be more involved than I thought on a first pass, so I'll just leave it for now and come back to that later. Focus on making the debs work properly first. 

The idea will be the dockerfile will fetch the latest build debs and install them, so the debian libindi version won't be a factor.

TN-1:
Manual builds can be done with: `act workflow_dispatch -W .github/workflows/package-deb.yml -j build-deb --matrix distro:trixie --matrix arch:amd64 --container-options "--privileged"` for some distro and arch that is supported. Needs this to be installed: https://github.com/nektos/act

TN-1:
My testing is progressing well. I have confirmed that the following are working:

Fresh deb installs
setup.sh > deb migration
In place upgrade with deb
Nginx instead of apache

Still need to test on ARM64 hardware, or with actual cameras

Paging @GuLinux @jerryfmedeiros @tswfi @ikeysolomon @EthanChappel @321MathiasS123 @jaxzin in case any of you would be willing to help test, find edge cases or bugs. Please backup, or run on test systems :)

321MathiasS123:
I think I have a dead Pi4 lying around somewhere, but I can't say for sure whether that's due to the SD-card, the power supply or the whole Pi being fried. I'll have to look into that. I don't have any suitable camera here outside of my production system and if that Pi is truly dead, any testing would have to take place in a VM. I'll see what I can do, but it may take a while.

debian/config
```diff
@@ -0,0 +1,278 @@
...
+            db_input high indi-allsky/elevation || true
+            if db_go; then
+                db_get indi-allsky/latitude
+                LAT="$RET"
+                db_get indi-allsky/longitude
+                LON="$RET"
+
+                # Validate Latitude (-90 to 90)
+                if [ -n "$LAT" ]; then
+                    INVALID_LAT=$(python3 -c "try: lat=float('$LAT'); print('0' if -90.0 <= lat <= 90.0 else '1')
```
jaxzin: The coordinate validator fails open: the `2>/dev/null || echo "0"` fallback means any value that breaks Python parsing (a stray quote, newline, backslash) is silently accepted as a **valid** latitude, so a malformed coordinate misconfigures the observatory location with no warning. Passing `$LAT` out-of-band (below) also avoids interpolating it into the `python3 -c` source.

```sh
INVALID_LAT=$(python3 -c 'import sys
try:
    v = float(sys.argv[1])
    print("0" if -90.0 <= v <= 90.0 else "1")
except Exception:
    print("1")
' "$LAT") || INVALID_LAT=1
```

debian/config
```diff
@@ -0,0 +1,278 @@
...
+                        db_fset indi-allsky/invalid_latitude seen false || true
+                        db_input critical indi-allsky/invalid_latitude || true
+                        db_go || true
+                        continue
+                    fi
+                fi
+
+                # Validate Longitude (-180 to 180)
+                if [ -n "$LON" ]; then
+                    INVALID_LON=$(python3 -c "try: lon=float('$LON'); print('0' if -180.0 <= lon <= 180.0 else '1')
```
jaxzin: Same root-injection + fail-open pattern as the latitude validator above (`$LON` interpolated into `python3 -c` source, `|| echo "0"` fails open). Apply the same `argv`-based, fail-closed fix here.

debian/postinst
```diff
@@ -0,0 +1,547 @@
...
+        # Only provision a default admin account on fresh installs (check if active admin user already exists)
+        DB_ADMIN_COUNT=0
+        if [ -f "$DB_FILE" ] && command -v sqlite3 >/dev/null 2>&1; then
+            DB_ADMIN_COUNT=$(sqlite3 "$DB_FILE" "SELECT count(*) FROM user WHERE admin = 1 AND username != 'system';" 2>/dev/null || echo "0")
+        fi
+
+        if [ "${DB_ADMIN_COUNT:-0}" -eq 0 ] && [ -f "/usr/share/indi-allsky/misc/usertool.py" ]; then
+            ADMIN_USER="${WEB_USER:-admin}"
+            ADMIN_NAME="${WEB_NAME:-Admin User}"
+            ADMIN_PASS="${WEB_PASS:-adminpassword}"
```
jaxzin: On a headless install this can silently create the admin account with the public default `adminpassword` on a network-facing login: when `WEB_PASS` is empty or under 8 chars, both this line and the guard just below fall back to the literal default with no warning. (It's empty in the first place because the documented preseed recipe sets `web_pass` but not `web_pass_confirm`, so `debian/config`'s mismatch branch resets both fields.)

Fix — generate a random password instead of a fixed one, and tell the operator (replaces lines 293 and 296–298):

```sh
if [ -z "$WEB_PASS" ] || [ "${#WEB_PASS}" -lt 8 ]; then
    ADMIN_PASS="$(openssl rand -base64 12)"
    echo "indi-allsky: no valid admin password supplied; generated: $ADMIN_PASS" >&2
    echo "indi-allsky: log in and change it immediately." >&2
else
    ADMIN_PASS="$WEB_PASS"
fi
```

Separately, seed `web_pass_confirm` in the deployment docs (or copy `web_pass` → `web_pass_confirm` in `debian/config` when confirm is empty) so a valid preseeded password isn't discarded in the first place.

debian/indi-allsky.indiserver.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=INDI Server for Camera Control
+After=network.target
+
+[Service]
+Type=simple
+User=indi-allsky
+Group=indi-allsky
+EnvironmentFile=-/etc/indi-allsky/indiserver.env
+ExecStart=/usr/bin/indiserver -v -p 7624 $CCD_DRIVER $GPS_DRIVER
```
jaxzin: `ExecStart` hardcodes `-p 7624`, but the debconf INDI port is only written to `flask.json`/SQLite — never to `indiserver.env`. Pick any other port and the client dials it while indiserver stays on 7624 → silent "not reachable". (The old `service/indiserver.service` parameterized `%INDI_PORT%`, so this is a regression.)

In `postinst`, add the port to the env file:

```sh
INDI_PORT="${INDI_PORT:-7624}"
```

…and reference it here (the unit already expands `$CCD_DRIVER` via `EnvironmentFile=`, so this works the same way):

```ini
ExecStart=/usr/bin/indiserver -v -p ${INDI_PORT} $CCD_DRIVER $GPS_DRIVER
```

debian/postinst
```diff
@@ -0,0 +1,547 @@
...
+    "SQLALCHEMY_DATABASE_URI": "sqlite:////var/lib/indi-allsky/indi-allsky.sqlite",
+    "CAMERA_INTERFACE": "${CAMERA_INTERFACE}",
+    "CCD_DRIVER": "${CCD_DRIVER}",
+    "INDI_ALLSKY_AUTH_ALL_VIEWS": ${WEB_AUTH_ALL_VIEWS:-false},
+    "LOCAL_AUTH_ENABLE": true,
+    "OIDC_ENABLE": ${OIDC_ENABLE:-false}
+}
+EOFJ
+            fi
+            chown "$ALLSKY_USER:www-data" "$FLASK_JSON" 2>/dev/null || true
```
jaxzin: (nit) The `flask.json` chown/chmod pair is repeated in four places, and the final `chown -R indi-allsky:indi-allsky /etc/indi-allsky` (~line 474) overrides all of them anyway. Consolidating into one `fix_permissions()` helper called once at the end would drop the redundancy and give a single place to fix the unset `$ALLSKY_USER` (currently these `chown`s only set the group).

jaxzin (changes requested):
I haven't exercised the full path yet, but this is what I found from my initial review. I may have a second round to share after my local build pipeline finished this evening.

jerryfmedeiros:
I gave this a go on arm64 since that was still on the list. Not a Pi with a real camera unfortunately, just native aarch64 containers on an Apple Silicon Mac, Debian 12 and 13. I will test it on a Pi with a camera as soon as I can. 

The arm64 build is fine. dpkg-buildpackage went through clean, all 109 wheels came down as native aarch64 manylinux wheels, and the offline cache installed without falling back to PyPI once. qa_test.sh passes end to end, and PyIndi, cv2, astropy, scipy, rawpy, dbus and systemd all import in the venv. Preseeded lat/lon and driver settings landed correctly in both flask.json and the sqlite config.

A few things I ran into:

- Install comes to about 1.15GB

/usr/share/indi-allsky is 254MB, and 227MB of that is the wheels directory. The venv is another 891MB. The wheels just sit there after postinst is done. On a Pi SD card that's a lot of space doing nothing. Could delete them at the end of a successful install, or split them into their own package people can remove afterwards.

- $ALLSKY_USER and $ALLSKY_GROUP are never set in postinst

They get used on lines 143, 152, 304, 305, 323 and 330. So chown "$ALLSKY_USER:www-data" ends up as chown ":www-data", which only sets the group, and line 304's chown -R ":" is just invalid. It's all swallowed by 2>/dev/null || true so nothing visibly breaks, and line 474 sets the right ownership anyway. But those six lines aren't doing what they look like they're doing.

- Recursive chown on the image directory on every upgrade

Is this really necessary? Line 474 does chown -R indi-allsky:indi-allsky /var/www/html/allsky. That's the whole image archive. Anyone with a year of captures has 100k+ files and this walks every one of them on an SD card, on every single upgrade. It's also the one chown without || true, so under set -e a single unreadable file aborts the install.

debian/indi-allsky.indiserver.service
TN-1: Already address this locally, just havent pushed yet :) Good catch

debian/config
TN-1: Great point

debian/postinst
TN-1: Good point

TN-1:
Thanks to you both for your feedback, I very much appreciate it!

> Could delete them at the end of a successful install, or split them into their own package people can remove afterwards.

Great catch, I will add a cleanup step. I think splitting out would be a better long term fix, but for now I think a simple cleanup will suffice.

> Recursive chown on the image directory on every upgrade

You're right, this isnt needed.

> $ALLSKY_USER and $ALLSKY_GROUP are never set in postinst

Fixed :)

Thanks for testing on the ARM64 hardware, great to hear that it worked there as well.

> I think I have a dead Pi4 lying around somewhere, but I can't say for sure whether that's due to the SD-card, the power supply or the whole Pi being fried. I'll have to look into that. I don't have any suitable camera here outside of my production system and if that Pi is truly dead, any testing would have to take place in a VM. I'll see what I can do, but it may take a while.

I do have another Pi and HQ cam built into a film scanner, so worst case scenario I could test on that

aaronwmorris:
@jerryfmedeiros 


>     Recursive chown on the image directory on every upgrade
> 
> Is this really necessary?

The original setup.sh script serves multiple purposes:

1. Setup the software
2. Upgrade the software
3. Fix problems caused by the user (or by me)

There are quite a few steps in setup.sh that are there specifically to fix possible problems caused by the user.

TN-1:
> @jerryfmedeiros
> 
> > ```
> > Recursive chown on the image directory on every upgrade
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Is this really necessary?
> 
> The original setup.sh script serves multiple purposes:
> 
> 1. Setup the software
> 2. Upgrade the software
> 3. Fix problems caused by the user (or by me)
> 
> There are quite a few steps in setup.sh that are there specifically to fix possible problems caused by the user.

Might be worth moving all the repair functions into its own script that can be called if needed, rather than hardcoding them into the package. I think thats the cleaner, more canon way of fixing it. Said script could be installed so a user can call it like `sudo indi-allsky-repair`

If you agree, would you be able to indicate which sections of setup are purely repair actions and can be ignored in the deb/put into the repair script? 

(Could even add a prompt at the end of the upgrade asking if users want to run it just in case)

TN-1:
@aaronwmorris In regards to your comment elsewhere, can the requirements for the wheels being installed be slimmed back at all since we are pre-building? Sounds like some of the reqs are just for compilation.

ETA: keep in mind, that having indi-allsky packaged as a deb will avoid most of the issues that users cause resulting in healing actions. The vector for users mistakes will be reduced quite a bit.

TN-1:
I have added indi-allsky-ctl, a command wrapper that gets installed at /usr/bin that will allow the user to run the user facing commands in /misc without having to worry about venv, etc.    
   
```
INDI Allsky Command-Line Control & Maintenance Utility

Usage:
  indi-allsky-ctl <command> [options/arguments...]

User & Authentication Management:
  passwd / reset-pass      Reset or change password for a web user
  user-add                 Create a new web user account
  user-list                List all configured web users
  user-del                 Delete a web user account

Database & Storage:
  backup-db                Create an immediate database backup
  fix-perms                Restore standard ownership & permissions (as in setup.sh)
  expire-images            Run image retention expiration and storage purge
  rebuild-thumbnails       Regenerate missing image thumbnails
  convert-db               Convert database between SQLite and MySQL
  validate-db              Audit database schema and repair broken records

Hardware & Diagnostics:
  test-camera              Run diagnostic camera capture test
  list-cameras             List detected INDI cameras
  test-sensors             Probe and read connected hardware sensors
  support-info             Generate diagnostic report bundle for troubleshooting

Peripheral Configuration:
  setup-gps                Configure GPS / GPSD service
  setup-hotspot            Configure WiFi hotspot failover
  setup-rtc                Configure I2C hardware Real-Time Clock
  setup-usb-mount          Configure automatic USB flash drive automount

Service Control:
  status                   Show status of indi-allsky and gunicorn services
  restart                  Restart indi-allsky and gunicorn services
  logs                     Follow recent logs from systemd journal
```

TN-1:
Reviewing the deps:
```
    Depends: ${misc:Depends},
     ${shlibs:Depends},
     python3-pyindi-client,
     python3 (>= 3.9),
     python3-venv,
     python3-pip,
     ffmpeg,
     apache2 | nginx,
     sqlite3,
     jq,
     * whiptail,
     debconf,
     adduser,
     openssl,
     bc,
     procps,
     rsyslog,
     logrotate,
     cron,
    * git,
    * cpio,
     tzdata,
     locales,
     ca-certificates,
    * libcfitsio10 | libcfitsio9 | libcfitsio-dev,
    * libnova-0.16-0 | libnova-dev,
    * libopencv-core406 | libopencv-core406t64 | libopencv-dev,
    * libraw20 | libraw23t64 | libraw-dev,
     network-manager,
     udisks2
```

I have marked removal candidates with an *, justification:
Git: Not needed, no git checkouts now
cpio: dpkg handles the copying
-dev and libs: Not needed if not building
whiptail: not used

Any objections?

aaronwmorris:
> In regards to your comment elsewhere, can the requirements for the wheels being installed be slimmed back at all since we are pre-building? Sounds like some of the reqs are just for compilation.

@TN-1 Yes, if the wheels are included in the deb file, the packages can be trimmed.

The `opencv` packages (opencv and openblas) are probably unnecessary.  The opencv python wheel will contain all of the C libraries needed to run.

The image libraries may not be needed (libjpeg, libpng, libtiff, libraw, libcfitsio).  It just depends if the python wheel includes all its own libraries.

debian/templates
```diff
@@ -0,0 +1,206 @@
...
+Description: Install GPIO and hardware sensor Python modules?
+ Install Adafruit CircuitPython, Blinka, and hardware sensor drivers (BME280, BMP280, TSL2591, DHT, SHT3x, MLX90614, INA219, etc.) for local telemetry.
+
+Template: indi-allsky/install_optional_modules
+Type: boolean
+Default: false
+Description: Install optional Python modules (cloud storage & PyMySQL)?
+ Install cloud storage upload drivers (AWS S3, Google Cloud Storage, Apache Libcloud, YouTube upload) and PyMySQL support.
+
+
```
jaxzin: The package won't install on this revision. `debian/templates` ends with an extra blank line, which debconf treats as the start of another template with no content, so it rejects the whole file — `postinst` then exits 255 and dpkg leaves the package unconfigured. I hit this in a clean bookworm container: `apt install` fails with `Template #33 in ... does not contain a 'Template:' line`. (Introduced by 7a7de8b3.)

Remove the trailing blank line so the file ends right after the last stanza:

```
...and PyMySQL support.
```
(i.e. delete the blank lines 205–206; `debconf-loadtemplate` then loads it cleanly.)

jaxzin (changes requested):

debian/indi-allsky.postinst
```diff
@@ -0,0 +1,832 @@
...
+                echo "indi-allsky: Offline wheel installation failed (host Python ABI mismatch or missing wheel)."
+            fi
+        fi
+
+        if [ "$OFFLINE_OK" != "1" ] && [ -f "$REQ_FILE" ]; then
+            echo "indi-allsky: Attempting online dependency resolution via PyPI..."
+            # shellcheck disable=SC2086
+            "$VENV/bin/pip" install --no-cache-dir $REQ_ARGS || {
+                echo "indi-allsky: WARNING: Online dependency installation encountered errors. Virtualenv may be incomplete."
+            }
```
jaxzin: If this `pip install` fails, it only logs a warning, so the install claims to succeed but the venv is empty and no database or admin user gets created. I hit this in a bookworm container on my M1 Mac by installing without `indi-allsky-wheels` — the `pip install` failed because `pkg-config` isn't installed.

Fail the install instead — add after the `fi` on line 238:

```sh
if ! "$VENV/bin/python3" -c 'import PyIndi, flask, sqlalchemy' 2>/dev/null; then
    echo "indi-allsky: ERROR: virtualenv incomplete (offline cache mismatch and/or fallback failed); service will not start." >&2
    exit 1
fi
```

debian/control
```diff
@@ -0,0 +1,103 @@
...
+Description: Python 3 bindings for INDI library (PyIndi)
+ This package provides compiled Python 3 bindings for the INDI library
+ framework, allowing Python scripts and applications to interact with
+ astronomical hardware drivers and indiserver via the PyIndi module.
+
+Package: indi-allsky-wheels
+Architecture: any
+Section: python
+Priority: optional
+Depends: ${misc:Depends}
```
jaxzin: The wheels in this package only work with Python 3.11, but it has no Python dependency, so apt will claim to install it on Python 3.12+ (Ubuntu 24.04, Debian 13). On those systems pip can't use the bundled wheels, so you end up in the failure from my comment on `debian/postinst` — the install claims to succeed with an empty venv.

Declare the Python version the wheels were built for, so apt refuses up front with a clear error:

```
Depends: ${misc:Depends}, python3 (>= 3.11), python3 (<< 3.12)
```

debian/postrm
```diff
@@ -0,0 +1,43 @@
...
+fi
+
+case "$1" in
+    purge)
+        if [ -e /usr/share/debconf/confmodule ]; then
+            db_purge
+        fi
+
+        # Remove indi-allsky core data and configuration directories
+        rm -rf /var/lib/indi-allsky /etc/indi-allsky /var/log/indi-allsky
```
jaxzin: Purge is inconsistent: this removes the database, venv, logs, and config, but leaves `/var/www/html/allsky` (the captured images) — yet the database is user data too (dark-frame calibrations, image history). So purge currently neither fully cleans up nor fully protects user data — nothing flags that the database (dark frames, image history) is being deleted, while gigabytes of images are left behind.

Pick one consistent policy:

- **Standard purge** (matches Debian semantics — a user who wants to keep data uses `apt remove`, not `purge`): remove the images too.
  ```sh
  rm -rf /var/lib/indi-allsky /etc/indi-allsky /var/log/indi-allsky /var/www/html/allsky
  ```
- **Or protect user data**: preserve both the captures *and* the database on purge, and print where they are.

jaxzin (changes requested):
On the last revision that installed, I built the `.deb` and tested it in a Debian 12 (bookworm) container on my M1 Mac. The problems I found are in my inline comments; here's what worked:
- **Offline install works.** Since the whole point of the wheel cache is avoiding compiles and network on a Pi, I installed with the container's network disabled and the app imported cleanly.
- **Upgrades keep your data.** Installing a newer version over an existing install kept the admin account and `flask.json` unchanged.
- **`debian/tests/qa_test.sh` passes** all seven probes.

debian/templates
TN-1: Hmmm, Weird, I didnt get this on my end. Thanks for flagging, have corrected.

debian/indi-allsky.postinst
TN-1: Good catch, thankyou

debian/postrm
TN-1: Thanks for catching this

setup.sh
```diff
@@ -164,9 +164,23 @@ if which whiptail >/dev/null 2>&1; then
...
+echo "###                                                                          ###"
+echo "###  setup.sh is now deprecated in favor of official Debian packaging (.deb).###"
+echo "###                                                                          ###"
+echo "###  Recommended install / upgrade from GitHub Releases:                     ###"
+echo "###    1. Download the latest .deb for your architecture (arm64/amd64):       ###"
+echo "###       https://github.com/aaronwmorris/indi-allsky/releases/latest         ###"
+echo "###    2. Install via apt:                                                   ###"
+echo "###       sudo apt update && sudo apt install ./indi-allsky_*.deb             ###"
+echo "###                                                                          ###"
+echo "###  Migration Guide: docs/setup_to_deb_migration.md                         ###"
```
TN-1, line 178: Thoughts on the deprecation notice?

TN-1:
> > In regards to your comment elsewhere, can the requirements for the wheels being installed be slimmed back at all since we are pre-building? Sounds like some of the reqs are just for compilation.
> 
> @TN-1 Yes, if the wheels are included in the deb file, the packages can be trimmed.
> 
> The `opencv` packages (opencv and openblas) are probably unnecessary. The opencv python wheel will contain all of the C libraries needed to run.
> 
> The image libraries may not be needed (libjpeg, libpng, libtiff, libraw, libcfitsio). It just depends if the python wheel includes all its own libraries.

@aaronwmorris I got it down to the following, do you notice any potential issues?

Build Depends:
- debhelper-compat (= 13)
- dh-python
- python3-all
- python3-setuptools
- python3-venv
- python3-dev
- python3-pip
- python3-wheel
- libindi-dev
- libcurl4-gnutls-dev
- libssl-dev
- libdbus-1-dev
- libglib2.0-dev
- libsystemd-dev
- swig
- git

Runtime Depends:
- python3-pyindi-client
- python3 (>= 3.9)
- python3-venv
- python3-pip
- ffmpeg
- apache2 | nginx
- sqlite3
- jq
- debconf
- adduser
- openssl
- bc
- procps
- rsyslog
- logrotate
- cron
- tzdata
- locales
- ca-certificates
- network-manager
- udisks2
 
Removed Dependencies:
- libopencv-dev
- libopenblas-dev
- libraw-dev
- libcfitsio-dev
- libnova-dev
- libimath-dev
- libopenexr-dev
- libtiff-dev
- libpng-dev
- zlib1g-dev
- libopencv-core*
- libraw20 / libraw23t64
- libcfitsio9 / libcfitsio10
- libnova-0.16-0
- git (from runtime)
- cpio
- whiptail*

TN-1:
Have added support for a github pages hosted apt repo. Setup instructions added to the OP. Features a nice aesthetic landing page with setup instructions.

image

aaronwmorris:
Where does the package `python3-pyindi-client` come from?

setup.sh
jaxzin: Looks good. I guess I didn't realize this is a deprecation of setup.sh. No arguments against it, just assumed both would be supported (setup.sh for other distros or the masochists).

TN-1: I mean, Aaron can keep setup.sh around if he likes but I think we would be better off getting rid of it because then we have 2 codepaths to maintain and more likely for feature drift to happen. Just need him to clarify which direction he wants to go :)

TN-1:
> Where does the package `python3-pyindi-client` come from?

We build it ourselves. In debian/rules:6-26 you can see where it happens. It comes from https://github.com/indilib/pyindi-client.git

debian/postrm
```diff
@@ -0,0 +1,52 @@
...
+        # Detect custom image folder before removing configuration
+        CUSTOM_IMG=""
+        if [ -f "/etc/indi-allsky/flask.json" ] && command -v jq >/dev/null 2>&1; then
+            CUSTOM_IMG=$(jq -r '.INDI_ALLSKY_IMAGE_FOLDER // empty' /etc/indi-allsky/flask.json 2>/dev/null || true)
+        fi
+
+        # Remove indi-allsky core data, web assets, images, and configuration directories
+        rm -rf /var/lib/indi-allsky /etc/indi-allsky /var/log/indi-allsky /var/www/html/allsky
+        if [ -n "$CUSTOM_IMG" ] && [ "$CUSTOM_IMG" != "/var/www/html/allsky/images" ] && [ "$CUSTOM_IMG" != "/" ] && [ "$CUSTOM_IMG" != "/home" ] && [ "$CUSTOM_IMG" != "/var" ]; then
+            rm -rf "$CUSTOM_IMG" 2>/dev/null || true
```
jaxzin: This added `rm -rf "$CUSTOM_IMG"` block should be removed. Deleting a folder outside the default install location is [contrary to Debian policy §6.8](https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html#details-of-removal-and-or-configuration-purging): when `INDI_ALLSKY_IMAGE_FOLDER` points elsewhere (`/mnt/nas/allsky`, or a trailing-slash `/home/` that slips the blocklist), this recursively wipes a path the package doesn't own — possibly a mount. The unconditional `rm -rf .../var/www/html/allsky` above already removes everything purge should touch — that tree is package-owned, so purging it is standard (keep-your-data = `apt remove`).

debian/indi-allsky.install
```diff
@@ -0,0 +1,12 @@
...
+app.py usr/share/indi-allsky/
+config.py usr/share/indi-allsky/
+flask.json_template usr/share/indi-allsky/
+darks.py usr/share/indi-allsky/
+indi_allsky usr/share/indi-allsky/
+migrations usr/share/indi-allsky/
+misc usr/share/indi-allsky/
+html usr/share/indi-allsky/
+requirements usr/share/indi-allsky/
+service usr/share/indi-allsky/
```
jaxzin: Add `log usr/share/indi-allsky/` here, mirroring the existing `service` entry. The manifest currently omits the `log/` directory, so `/usr/share/indi-allsky/log/` is never packaged and the rsyslog/logrotate install blocks in `postinst` (lines 470/481) silently no-op — the README-promised log routing and daily rotation never happen.

misc/indi-allsky-ctl
```diff
@@ -0,0 +1,222 @@
...
+    DB_FOLDER="/var/lib/indi-allsky"
+    DB_FILE="${DB_FOLDER}/indi-allsky.sqlite"
+    VARLIB_FOLDER="/var/lib/indi-allsky"
+    LOG_FOLDER="/var/log/indi-allsky"
+    HTDOCS_FOLDER="/var/www/html/allsky"
+
+    # Resolve IMAGE_FOLDER dynamically from flask.json if present
+    IMAGE_FOLDER="${HTDOCS_FOLDER}/images"
+    if [ -f "${ALLSKY_ETC}/flask.json" ] && command -v jq >/dev/null 2>&1; then
+        CFG_IMAGE_FOLDER=$(jq -r '.IMAGE_FOLDER // empty' "${ALLSKY_ETC}/flask.json" 2>/dev/null || true)
```
jaxzin: Change line 72 to `jq -r '.INDI_ALLSKY_IMAGE_FOLDER // empty'`. It currently reads `.IMAGE_FOLDER`, but the actual key in flask.json is `INDI_ALLSKY_IMAGE_FOLDER`, so `fix-perms` always falls back to the default path and silently skips any custom image folder.

debian/rules
```diff
@@ -0,0 +1,42 @@
...
+
+%:
+	dh $@ --with python3
+
+override_dh_auto_build:
+	# Clone and compile pyindi-client for python3-pyindi-client package
+	mkdir -p debian/pyindi-build
+	if [ ! -f debian/pyindi-build/setup.py ]; then \
+		git clone --depth 1 https://github.com/indilib/pyindi-client.git debian/pyindi-build; \
+		(cd debian/pyindi-build && git checkout d8ad88f 2>/dev/null || true); \
```
jaxzin: Fetch the exact SHA (`git fetch --depth 1 origin d8ad88f && git checkout FETCH_HEAD`) and drop `|| true` so a failed pin aborts the build. As written, `git clone --depth 1` never fetches commit `d8ad88f`, so this `git checkout d8ad88f` fails and `2>/dev/null || true` swallows the error, leaving the build compiled against the moving upstream tip instead of the pinned commit.

.github/workflows/package-deb.yml
```diff
@@ -0,0 +1,514 @@
...
+            echo "### Bundled INDI & Driver Versions"
+            echo "* **INDI Core / Server**: \`${{ env.INDI_CORE_TAG }}\`"
+            echo "* **INDI 3rd-Party Drivers**: \`${{ env.INDI_3RDPARTY_TAG }}\` (ZWO ASI, QHY, Player One, ToupTek, SVBONY, Starlight Xpress, etc.)"
+            echo "* **Supported Architectures**: \`arm64\` (Raspberry Pi 4/5, aarch64), \`amd64\` (x86_64)"
+            echo "* **Distributions**: Debian 12 (Bookworm), Debian 13 (Trixie), Ubuntu 24.04 (Noble)"
+            echo ""
+            echo "### What's Changed"
+            # Ensure all git tags are fetched locally
+            git fetch --tags --force 2>/dev/null || true
+            PREV_TAG=$(git tag -l "indi_v*" | sort -V | tail -n 1)
```
jaxzin: Exclude the current ref before selecting: `... | sort -V | grep -vx "${GITHUB_REF_NAME}" | tail -n 1`. On a tagged release HEAD is the pushed tag, so the current `PREV_TAG=... | tail -n 1` resolves to that same tag and `git log "${PREV_TAG}..HEAD"` yields an empty range — publishing a release body with no changelog.

debian/postinst
```diff
@@ -0,0 +1,839 @@
...
+            PASSWORD_KEY=$(python3 -c "from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())" 2>/dev/null || python3 -c "import base64, os; print(base64.urlsafe_b64encode(os.urandom(32)).decode())")
+
+            if [ -f "$FLASK_TEMPLATE" ] && command -v jq >/dev/null 2>&1; then
+                # Start from the full template and substitute debconf values
+                jq \
+                    --arg secret_key "$SECRET_KEY" \
+                    --arg password_key "$PASSWORD_KEY" \
+                    --arg camera_interface "${CAMERA_INTERFACE}" \
+                    --arg ccd_driver "${CCD_DRIVER}" \
+                    --argjson http_port "${HTTP_PORT:-80}" \
```
jaxzin: Validate the port inputs in `debian/config` (like the latitude/longitude checks), or write the jq output to a temp file and `mv` it into place only on success. Otherwise a non-numeric debconf port answer makes this unguarded `jq --argjson` fail under `set -e` after the `>` redirect has already truncated flask.json — aborting `configure` and leaving a 0-byte config with a half-installed package.

misc/bump_version.py
```diff
@@ -0,0 +1,95 @@
...
+        data = json.loads(PACKAGE_JSON.read_text())
+        data["version"] = clean_ver
+        PACKAGE_JSON.write_text(json.dumps(data, indent=2) + "\n")
+        print(f" Updated {PACKAGE_JSON.relative_to(ROOT_DIR)}")
+
+    # 3. Update base.html CSS cache-buster default fallback
+    if BASE_HTML.exists():
+        html_content = BASE_HTML.read_text()
+        html_content = re.sub(
+            r"(\?v=(?:{{[^}]+}}|'))v\d{4}\.\d{2}\.\d+[^'\"]*",
```
jaxzin: Target the `else`-clause literal — `r"(else ')v\d{4}\.\d{2}\.\d+[^']*"` — and print "Updated" only when the `re.subn` count is non-zero. As written, this regex's `{{[^}]+}}` alternative consumes the whole Jinja block, so it never matches the version literal nested inside `{{ release if release else '...' }}` and the cache-buster is silently never bumped (while still printing "Updated").

debian/postrm
jaxzin: This added `rm -rf "$CUSTOM_IMG"` block should be removed. When `INDI_ALLSKY_IMAGE_FOLDER` points outside the default tree (`/mnt/nas/allsky`, or a trailing-slash `/home/` that slips the blocklist), a routine `apt purge` recursively deletes a directory the package never created — potentially a mount point, taking the whole share with it. The unconditional `rm -rf .../var/www/html/allsky` above already removes the package's own image tree, which is all purge should touch; a location the admin chose for their data isn't the package's to delete.

debian/indi-allsky.install
jaxzin: Add `log usr/share/indi-allsky/` here, mirroring the existing `service` entry. The manifest currently omits the `log/` directory, so `/usr/share/indi-allsky/log/` is never packaged and the rsyslog/logrotate install blocks in `postinst` (lines 470/481) silently no-op — the README-promised log routing and daily rotation never happen.

misc/indi-allsky-ctl
jaxzin: Change line 72 to `jq -r '.INDI_ALLSKY_IMAGE_FOLDER // empty'`. It currently reads `.IMAGE_FOLDER`, but the actual key in flask.json is `INDI_ALLSKY_IMAGE_FOLDER`, so `fix-perms` always falls back to the default path and silently skips any custom image folder.

debian/rules
jaxzin: Fetch the exact SHA (`git fetch --depth 1 origin d8ad88f && git checkout FETCH_HEAD`) and drop `|| true` so a failed pin aborts the build. As written, `git clone --depth 1` never fetches commit `d8ad88f`, so this `git checkout d8ad88f` fails and `2>/dev/null || true` swallows the error, leaving the build compiled against the moving upstream tip instead of the pinned commit.

.github/workflows/package-deb.yml
jaxzin: Exclude the current ref before selecting: `... | sort -V | grep -vx "${GITHUB_REF_NAME}" | tail -n 1`. On a tagged release HEAD is the pushed tag, so the current `PREV_TAG=... | tail -n 1` resolves to that same tag and `git log "${PREV_TAG}..HEAD"` yields an empty range — publishing a release body with no changelog.

debian/postinst
jaxzin: Validate the port inputs in `debian/config` (like the latitude/longitude checks), or write the jq output to a temp file and `mv` it into place only on success. Otherwise a non-numeric debconf port answer makes this unguarded `jq --argjson` fail under `set -e` after the `>` redirect has already truncated flask.json — aborting `configure` and leaving a 0-byte config with a half-installed package.

misc/bump_version.py
jaxzin: Target the `else`-clause literal — `r"(else ')v\d{4}\.\d{2}\.\d+[^']*"` — and print "Updated" only when the `re.subn` count is non-zero. As written, this regex's `{{[^}]+}}` alternative consumes the whole Jinja block, so it never matches the version literal nested inside `{{ release if release else '...' }}` and the cache-buster is silently never bumped (while still printing "Updated").

aaronwmorris:
While I was reading about the python module, I recalled that there is a prebuilt pyindi-client on PyPI.  https://pypi.org/project/pyindi-client/

Also, I am expecting that the INDI project may soon provide a Debian APT repo for INDI.  It has been proposed about a month ago:  https://github.com/indilib/indi/issues/2418

debian/postrm
jaxzin: This added `rm -rf "$CUSTOM_IMG"` block should be removed. When `INDI_ALLSKY_IMAGE_FOLDER` points outside the default tree (`/mnt/nas/allsky`, or a trailing-slash `/home/` that slips the blocklist), a routine `apt purge` recursively deletes a directory the package never created — potentially a mount point, taking the whole share with it. The unconditional `rm -rf .../var/www/html/allsky` above already removes the package's own image tree, which is all purge should touch; a location the admin chose for their data isn't the package's to delete.

debian/indi-allsky.install
jaxzin: Add `log usr/share/indi-allsky/` here, mirroring the existing `service` entry. The manifest currently omits the `log/` directory, so `/usr/share/indi-allsky/log/` is never packaged and the rsyslog/logrotate install blocks in `postinst` (lines 470/481) silently no-op — the README-promised log routing and daily rotation never happen.

misc/indi-allsky-ctl
jaxzin: Change line 72 to `jq -r '.INDI_ALLSKY_IMAGE_FOLDER // empty'`. It currently reads `.IMAGE_FOLDER`, but the actual key in flask.json is `INDI_ALLSKY_IMAGE_FOLDER`, so `fix-perms` always falls back to the default path and silently skips any custom image folder.

debian/rules
jaxzin: Fetch the pin by its full commit SHA and drop `2>/dev/null || true` so a bad pin fails the build loudly — pinning an immutable SHA (not a tag, which upstream can re-point) is the supply-chain-safe choice. I'd bump the pin to **v2.0.4**, the first pyindi-client release containing this change (`d8ad88f` is 3 commits before it), and annotate the SHA with the version for readability:

```make
# pyindi-client v2.0.4
PYINDI_CLIENT_SHA = 6f1f9ee0a449d0a64b0d755ac5fd8308ec322cae
...
git init debian/pyindi-build && cd debian/pyindi-build && \
  git remote add origin https://github.com/indilib/pyindi-client.git && \
  git fetch --depth 1 origin $(PYINDI_CLIENT_SHA) && \
  git checkout FETCH_HEAD
```

As written, `git clone --depth 1` fetches only master's tip (currently a year newer than `d8ad88f`), so `git checkout d8ad88f` fails, `|| true` swallows it, and the build silently compiles upstream tip instead of the pin.

.github/workflows/package-deb.yml
jaxzin: Exclude the current ref before selecting: `... | sort -V | grep -vx "${GITHUB_REF_NAME}" | tail -n 1`. On a tagged release HEAD is the pushed tag, so the current `PREV_TAG=... | tail -n 1` resolves to that same tag and `git log "${PREV_TAG}..HEAD"` yields an empty range — publishing a release body with no changelog.

debian/postinst
jaxzin: Validate the port inputs in `debian/config` (like the latitude/longitude checks), or write the jq output to a temp file and `mv` it into place only on success. Otherwise a non-numeric debconf port answer makes this unguarded `jq --argjson` fail under `set -e` after the `>` redirect has already truncated flask.json — aborting `configure` and leaving a 0-byte config with a half-installed package.

misc/bump_version.py
jaxzin: Target the `else`-clause literal — `r"(else ')v\d{4}\.\d{2}\.\d+[^']*"` — and print "Updated" only when the `re.subn` count is non-zero. As written, this regex's `{{[^}]+}}` alternative consumes the whole Jinja block, so it never matches the version literal nested inside `{{ release if release else '...' }}` and the cache-buster is silently never bumped (while still printing "Updated").

TN-1:
> While I was reading about the python module, I recalled that there is a prebuilt pyindi-client on PyPI.  https://pypi.org/project/pyindi-client/
> 
> Also, I am expecting that the INDI project may soon provide a Debian APT repo for INDI.  It has been proposed about a month ago:  https://github.com/indilib/indi/issues/2418

Oh nice, that explains why all the work has been done for the debian packaging. Once their repo is live, then I can just adjust our action to pull from their instead of building it ourselves. 

I will update to use pyindi from pypi 

Edit: had a look at the linked issue, looks like they are only building for Trixie whereas we are doing Bookworm and Noble too, so unless we want to cut back support we will still be building.

debian/postrm
jaxzin: This added `rm -rf "$CUSTOM_IMG"` block should be removed. When `INDI_ALLSKY_IMAGE_FOLDER` points outside the default tree (`/mnt/nas/allsky`, or a trailing-slash `/home/` that slips the blocklist), a routine `apt purge` recursively deletes a directory the package never created — potentially a mount point, taking the whole share with it. The unconditional `rm -rf .../var/www/html/allsky` above already removes the package's own image tree, which is all purge should touch; a location the admin chose for their data isn't the package's to delete.

debian/indi-allsky.install
jaxzin: Add `log usr/share/indi-allsky/` here, mirroring the existing `service` entry. The manifest currently omits the `log/` directory, so `/usr/share/indi-allsky/log/` is never packaged and the rsyslog/logrotate install blocks in `postinst` (lines 470/481) silently no-op — the README-promised log routing and daily rotation never happen.

misc/indi-allsky-ctl
jaxzin: Change line 72 to `jq -r '.INDI_ALLSKY_IMAGE_FOLDER // empty'`. It currently reads `.IMAGE_FOLDER`, but the actual key in flask.json is `INDI_ALLSKY_IMAGE_FOLDER`, so `fix-perms` always falls back to the default path and silently skips any custom image folder.

debian/rules
jaxzin: Fetch the pinned commit directly and check it out, and drop `2>/dev/null || true` so a broken pin fails the build instead of silently passing:

```make
git init debian/pyindi-build && cd debian/pyindi-build && \
  git remote add origin https://github.com/indilib/pyindi-client.git && \
  git fetch --depth 1 origin d8ad88fd7b0b55e801d52444e08d1cd625413561 && \
  git checkout FETCH_HEAD
```

`git clone --depth 1` only pulls master's tip, so `d8ad88f` isn't in the shallow clone and the current `git checkout d8ad88f` fails — swallowed by `|| true`, leaving the build on upstream tip. The fetch has to use the full 40-char SHA: git's `want` protocol doesn't accept abbreviated hashes, which is why the short `d8ad88f` works for a local `checkout` but can't be `fetch`ed.

debian/rules
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/make -f
+
+%:
+	dh $@ --with python3
+
+override_dh_auto_build:
+	# Clone and compile pyindi-client for python3-pyindi-client package
+	mkdir -p debian/pyindi-build
+	if [ ! -f debian/pyindi-build/setup.py ]; then \
+		git clone --depth 1 https://github.com/indilib/pyindi-client.git debian/pyindi-build; \
```
jaxzin: Optional, separate from the fix above: if you'd prefer a human-readable pin, you could point it at a tagged release instead of a bare commit — `v2.0.4` (`6f1f9ee`, first release with this change) or `v2.1.2` (`b320c5a`, latest). Still fetch by full SHA for immutability; the version just goes in a comment for readability.

.github/workflows/package-deb.yml
jaxzin: Exclude the current ref before selecting: `... | sort -V | grep -vx "${GITHUB_REF_NAME}" | tail -n 1`. On a tagged release HEAD is the pushed tag, so the current `PREV_TAG=... | tail -n 1` resolves to that same tag and `git log "${PREV_TAG}..HEAD"` yields an empty range — publishing a release body with no changelog.

debian/postinst
jaxzin: Validate the port inputs in `debian/config` (like the latitude/longitude checks), or write the jq output to a temp file and `mv` it into place only on success. Otherwise a non-numeric debconf port answer makes this unguarded `jq --argjson` fail under `set -e` after the `>` redirect has already truncated flask.json — aborting `configure` and leaving a 0-byte config with a half-installed package.

misc/bump_version.py
jaxzin: Target the `else`-clause literal — `r"(else ')v\d{4}\.\d{2}\.\d+[^']*"` — and print "Updated" only when the `re.subn` count is non-zero. As written, this regex's `{{[^}]+}}` alternative consumes the whole Jinja block, so it never matches the version literal nested inside `{{ release if release else '...' }}` and the cache-buster is silently never bumped (while still printing "Updated").

TN-1:
Have updated the flow to get pyindi-client from pypi and build it into the wheels deb

debian/postrm
jaxzin: This added `rm -rf "$CUSTOM_IMG"` block should be removed. When `INDI_ALLSKY_IMAGE_FOLDER` points outside the default tree (`/mnt/nas/allsky`, or a trailing-slash `/home/` that slips the blocklist), a routine `apt purge` recursively deletes a directory the package never created — potentially a mount point, taking the whole share with it. The unconditional `rm -rf .../var/www/html/allsky` above already removes the package's own image tree, which is all purge should touch; a location the admin chose for their data isn't the package's to delete.

debian/indi-allsky.install
jaxzin: Add `log usr/share/indi-allsky/` here, mirroring the existing `service` entry. The manifest currently omits the `log/` directory, so `/usr/share/indi-allsky/log/` is never packaged and the rsyslog/logrotate install blocks in `postinst` (lines 470/481) silently no-op — the README-promised log routing and daily rotation never happen.

misc/indi-allsky-ctl
jaxzin: Change the inner command to `jq -r '.INDI_ALLSKY_IMAGE_FOLDER // empty'`. It currently reads `.IMAGE_FOLDER`, but the actual key in flask.json is `INDI_ALLSKY_IMAGE_FOLDER`, so `fix-perms` always falls back to the default path and silently skips any custom image folder.

.github/workflows/package-deb.yml
jaxzin: Exclude the current ref before selecting: `... | sort -V | grep -vx "${GITHUB_REF_NAME}" | tail -n 1`. On a tagged release HEAD is the pushed tag, so the current `PREV_TAG=... | tail -n 1` resolves to that same tag and `git log "${PREV_TAG}..HEAD"` yields an empty range — publishing a release body with no changelog.

debian/postinst
jaxzin: Validate the port inputs in `debian/config` (like the latitude/longitude checks), or write the jq output to a temp file and `mv` it into place only on success. Otherwise a non-numeric debconf port answer makes this unguarded `jq --argjson` fail under `set -e` after the `>` redirect has already truncated flask.json — aborting `configure` and leaving a 0-byte config with a half-installed package.

misc/bump_version.py
jaxzin: Target the `else`-clause literal — `r"(else ')v\d{4}\.\d{2}\.\d+[^']*"` — and print "Updated" only when the `re.subn` count is non-zero. As written, this regex's `{{[^}]+}}` alternative consumes the whole Jinja block, so it never matches the version literal nested inside `{{ release if release else '...' }}` and the cache-buster is silently never bumped (while still printing "Updated").

jaxzin (changes requested):

debian/postinst
TN-1: Good catch, thanks

misc/indi-allsky-ctl
TN-1: Doh, nice catch

debian/postrm
TN-1: Yeah good call

TN-1:
Ive got rid of the wheels package, we just install the venv directly now. Its faster, and nicer syntax wise. Less work in postinst too.

TN-1:
The APT repo on my fork is now live for testing: https://tn-1.github.io/indi-allsky/

aaronwmorris:
Long term, I want to avoid forcing the user from upgrading the indilib libraries.  One of the things I had to add to the upgrade documentation is recommend NOT upgrading INDI unless they had a specific reason for doing so.

Upgrading indilib commonly introduces breaking changes.  The API/interface is stable, but there are regularly changes that causing irritations...

Specifically:

- Camera name changes.  Changes to the vendor SDK can change the reported camera name.  indi-allsky uses the camera name as the unique key for the camera in its database
- Server name changes.  This happens less often, but it does happen.  Historically, `indi_sv305_ccd` was renamed `indi_svbony_ccd` and then renamed again `indi_svbonycam_ccd`.  `indi_altair_ccd` was renamed `indi_altaircam_ccd`

jimmauck:
Aaron, I’d bet you have thought of this but is it possible to separate physical-device identity from transient INDI identifiers?

aaronwmorris:
> Aaron, I’d bet you have thought of this but is it possible to separate physical-device identity from transient INDI identifiers?

Only recently has it become a possibility.  In more recent versions of indilib, SOME (not all) cameras have begun exposing an optional serial number field which could be used uniquely identify a camera.  The serial number attribute is not standardized across cameras.

- ZWO cameras consistently have a good serial number attribute
- PlayerOne appears to expose a serial number (this driver was based on the ZWO code)
- My test Altair GPCAM 224C shows a serial number
- I have a Touptek microscope camera that has the same serial number attribute as Altair (these are in the same driver family) but the serial number is blank :-)

My Touptek microscope camera is a fun one as well... it is a `IUC26000KPA` which is a IMX571 APS-C (same as ASI2600) sensor.  This camera randomly switches between two different names, `IUC26000KPA` when it negotiates as USB 3.0 and `IUC26000KPA(USB2.0)` which in USB 2.0 mode.  What causes it to switch?  Nobody knows.  :-)

If you look at the database model for the camera table, this is why there are `name`, `name_alt1`, and `name_alt2` fields.  This allows multiple names to be configured for the same camera.

I can start working on a way to integrate the serial number in the device discovery.

TN-1:
> Long term, I want to avoid forcing the user from upgrading the indilib libraries. One of the things I had to add to the upgrade documentation is recommend NOT upgrading INDI unless they had a specific reason for doing so.
> 
> Upgrading indilib commonly introduces breaking changes. The API/interface is stable, but there are regularly changes that causing irritations...
> 
> Specifically:
> 
> * Camera name changes.  Changes to the vendor SDK can change the reported camera name.  indi-allsky uses the camera name as the unique key for the camera in its database
> * Server name changes.  This happens less often, but it does happen.  Historically, `indi_sv305_ccd` was renamed `indi_svbony_ccd` and then renamed again `indi_svbonycam_ccd`.  `indi_altair_ccd` was renamed `indi_altaircam_ccd`

I believe this is already the case. The workflow has pinned indi-core and indi-3rdparty versions, and manual runs can target specific versions to test before bumping it for everyone. In package-deb.yml:21:
```
env:
  INDI_CORE_TAG: 'v2.2.4.2'
  INDI_3RDPARTY_TAG: 'v2.2.4.1'
```

Is that what you meant or did you have something else in mind?

aaronwmorris:
The point is not forcing individual users to upgrade INDI.  Once a camera is in a working state with INDI, forcing users to upgrade can break their system.

TN-1:
Ah, I get you now. So indi-bin and indi-3rdparty are Recommends, not hard Depends, so indi-allsky can be upgraded separately to indi-bin, etc.    

The scenarios that would upgrade indi-bin and indi-3rdparty are:
Full system `apt upgrade`
apt install with new indi-bin debs

This can be mitigated via apt by `apt-mark hold indi-bin indi-3rdparty`, this will pin those 2 packages so apt will not touch them. We can document this, and/or add something to `indi-allsky-ctl` to help users do this.

Edit: @aaronwmorris We could add a question to the debconf asking if users want to hold the indi versions, thoughts?

TN-1:
@aaronwmorris Have added notes to the migration guide for the various INDI holding scenarios, thoughts?

## Special Scenarios & INDI Driver Management

### Scenario A: Preserving Source-Built INDI (Custom Builds)
If you previously compiled INDI / drivers from source (e.g., using `setup.sh`'s source build or manual git builds) and want to ensure APT does not overwrite your custom binaries:

#### Option 1: Protect Before Install (from Existing Git Checkout)
From your existing `indi-allsky` git clone directory:
```bash
sudo bash misc/indi-allsky-ctl protect-source-indi
sudo apt install -y indi-allsky
```

#### Option 2: Lean Initial Install + Post-Install Protection
If you don't have the git checkout handy:
```bash
# 1. Install without downloading recommended driver packages
sudo apt install --no-install-recommends -y indi-allsky

# 2. Register source build so all future 'apt upgrade' runs are safe
sudo indi-allsky-ctl protect-source-indi
```

Both options register a local `indi-local-source` dummy package (version `99.0.0`) in the `dpkg` database that satisfies all INDI dependencies. APT will recognize your source build and will never overwrite your binaries in `/usr/bin` or `/usr/lib`.

---

### Scenario B: Pinning (Holding) Installed APT INDI Driver Versions
If you install INDI drivers via APT but want to guarantee that full system upgrades (`sudo apt upgrade`) do not touch or update your camera drivers:

1. **Pin (Hold) Drivers**:
   ```bash
   sudo indi-allsky-ctl hold-indi
   ```
2. **Check Pin Status**:
   ```bash
   indi-allsky-ctl status-indi-hold
   ```
3. **Unpin (Allow Upgrades)**:
   ```bash
   sudo indi-allsky-ctl unhold-indi
   ```

---

### Scenario C: Raspberry Pi HQ / Camera Module (Lean Install)
If you are using a native Raspberry Pi camera (`rpicam-apps` / `libcamera`) and do not want to download the 3rd-party USB camera driver bundle:
```bash
sudo apt install --no-install-recommends -y indi-allsky rpicam-apps
```

aaronwmorris:
Putting a hold on the libindi package (or one of the base packages) was pretty much my plan.

TN-1:
image

Apt repo is now at: https://apt.indi-allsky.org/ 

It is tested and working. 

Hosting on Github pages didnt end up working, its now in a Cloudflare R2 bucket. Free tier should be enough for our needs.

I have also migrated my production Allsky Camera (Pi with HQ Camera) to a deb install with no issues. If someone with indi-3rdparty drivers could try migrating to a deb install thatd be good 

I have also split the web server out into indi-allsky-web, allowing for Web Only installs to be possible. (indi-allsky pulls in -web as a depends to allow a full install)

I think we are on the home stretch with this one

TN-1:
@aaronwmorris Is there any pressing need to run a vacuum on every upgrade? No issues keeping it, it just presents a big blocker in the apt flow. Just looking for optimisations at this point, am just curious how needed it actually is?

aaronwmorris:
It is not necessary, I just added as a best practice.  

Once a camera has been running for more than a few months, the database will continuously grow in size.  The only negative consequence of not running it is the space usage.

TN-1:
Sweet. For the deb, I have added a tracking file so a vacuum is only done if it has been more than 30 days since the previous. So users that track the nightly branch dont have to wait x minutes per upgrade for a vacuum.

TN-1:
This is going well, I have been using the APT repo on my production camera for about a week now and no issues, daily build upgrades have been going well.

At this point I would call this stable enough for others to test. Hoping for someone not using libcamera to do a test in particular. I have done test installs with a simulator on Ubuntu 24.06, 26.06, Debian 12 and 13, Pi OS 13.

I have also added support for 26.04 as suggested in #3118 

Does anyone have any final blockers identified?

TN-1:
@aaronwmorris This is ready. Before you merge, Ill need to email you the deployment secrets. After merging, it would be good to do a 2026.8.2 release to populate the stable release channel.

aaronwmorris:
Very nice.  I am looking through things now.

aaronwmorris:
Can I get the `MIGRATION_FOLDER` in the flask.json to be set to the actual folder being used?  I would much prefer to have the setting used rather than an assumed config.

Does this change rely on any other PRs, like the systemd service unit change?

TN-1:
> Can I get the `MIGRATION_FOLDER` in the flask.json to be set to the actual folder being used? I would much prefer to have the setting used rather than an assumed config.

That is the case, but there was some unneeded fallback logic I have trimmed.

> Does this change rely on any other PRs, like the systemd service unit change?

Nope, it is self contained. The service files used for the deb already have the changes made in #3111, That PR just backports the changes so setup.sh can use it too.

TN-1:
I have added a line to support info to reflect the installation method: `Install: deb (APT channel: nightly, package: indi-allsky 2026.08.2~nightly.20260911.1158+trixie)` Just under the system, and just above the hostname:  
  
```
#################################
### indi-allsky support info  ###
#################################

Distribution: debian
Release: 13
Arch: aarch64
Bits: 64

CPUs: 4
Memory: 3707036 kB

System: Raspberry Pi 4 Model B Rev 1.5 
Install: deb (APT channel: nightly, package: indi-allsky 2026.08.2~nightly.20260911.1158+trixie)

Linux srv1-allsky.hamishwest.xyz 6.18.39+rpt-rpi-v8 #1 SMP PREEMPT Debian 1:6.18.39-1+rpt1 (2026-07-29) aarch64 GNU/Linux
Hostname
srv1-allsky
srv1-allsky.hamishwest.xyz
 Static hostname: srv1-allsky.hamishwest.xyz
```

aaronwmorris:
I noticed the fallback logic for `MIGRATION_FOLDER`, but I tested the APT package on a lab machine and the migration folder is set to `/var/lib/indi-allsky/migrations` in flask.json , but that folder does not exist so it went to the fallback config.

TN-1:
> I noticed the fallback logic for `MIGRATION_FOLDER`, but I tested the APT package on a lab machine and the migration folder is set to `/var/lib/indi-allsky/migrations` in flask.json , but that folder does not exist so it went to the fallback config.

Was this a fresh install or a migration from a setup.sh install?

aaronwmorris:
Fresh install

TN-1:
I have hardened up the creation of directories and their permissions which should fix it. If you do an `apt purge` and then install again it should work as expected. Just need to wait for the action to finish to update the repo: https://github.com/TN-1/indi-allsky/actions/runs/34611403274

TN-1:
Fairly sure I found the issue, I have hardened up the creation of directories and their permissions which should fix it. If you do an apt purge and then install again it should work as expected. Just need to wait for the action to finish to update the repo: https://github.com/TN-1/indi-allsky/actions/runs/34613036580/job/103308079598

TN-1:
Build is done, ready to test. Issue should be fixed

Edit: Tested, working as expected now.

aaronwmorris:
I have the apt repo setup on a fresh VM, but I do not see the `indi-allsky` package anymore

TN-1:
Ensure you are adding the nightly channel. I updated the apt repo page for this PR but there is no stable release up yet :)

aaronwmorris:
Duh, the page was set to stable.  I did not notice.

TN-1:
Yeah sorry about that :) It was set to nightly by default, but I changed it to be stable by default getting ready to merge.

aaronwmorris:
I am still getting the wrong migration folder on a new build.

package version `2026.08.2~nightly.20260911.1651+resolute`

```
  "MIGRATION_FOLDER": "/var/lib/indi-allsky/migrations",
```

Edit:  Or is the `/var/lib/indi-allsky/migrations` folder the correct path?

TN-1:
> I am still getting the wrong migration folder on a new build.
> 
> package version `2026.08.2~nightly.20260911.1651+resolute`
> 
> ```
>   "MIGRATION_FOLDER": "/var/lib/indi-allsky/migrations",
> ```
> 
> Edit:  Or is the `/var/lib/indi-allsky/migrations` folder the correct path?

`/var/lib/indi-allsky/migrations` is correct 

`/var/lib/indi-allsky` is for things that change, `/usr/share/indi-allsky` is for things that dont change.

Having migrations in `/usr/share` was a mistake, and the fallback code wasnt needed at all.

aaronwmorris:
@TN-1 I am basically ready to merge this.  You can email me the deployment secret at aaron@aarmor.net and I will get this going.

TN-1:
@aaronwmorris Sent them through

TN-1:
@aaronwmorris Just FYI: If the deployment gets stuck like this, you can cancel the task (Not the whole action) and start it again. 

Not sure what causes it, seems to be pretty random, I assume it is GH infrastructure issues.

https://github.com/aaronwmorris/indi-allsky/actions/runs/34801241713/job/103844217221

aaronwmorris:
Thanks for the pointer!

TN-1:
@aaronwmorris I think the failure is an issue with the keys, note:

```
AWS_ACCESS_KEY_ID: ***
AWS_SECRET_ACCESS_KEY:  ***
```

Looks like an extra space got slipped into the start of the secret key. Can you try copying it in again?

Note: Despite the logs using AWS key names, the R2_ key names I sent you are correct. They get translated by the R2 API so the logs show a different key title, but they are the same keys.

Once the key is fixed, you can just run that task again and it should work ok.

Looks like the release notes worked, Havent been able to test that until now so yay :partying_face:

PR #3081, opened 2026-08-23
fix sidebar scaling

TN-1:
#3054 didnt update dist.css, so the resizing changes dont work as expected. This PR adjusts the scaling a pinch and rebuilds dist.css.


aaronwmorris:
I think it is the `dist.css` that is constantly causing a merge conflict.  The last few PRs I have merged, I had to manually fix and I may have caused this problem.

PR #3111, opened 2026-08-30
Refactor: Replace service timers with explicit unit dependencies

TN-1:
Replaces the arbitrary delay timers (`indiserver.timer` 30s delay, `indi-allsky.timer` 2min delay) with standard systemd unit ordering and dependency chains (`After=` and `Wants=`). This change is from my deb packaging work but I felt it was worth backporting


aaronwmorris:
@TN-1 I am fine with this.  This was originally how things were configured, but I added the timers a couple of years ago.  I no longer recall the full justification for this change.

The service control function for the indiserver and indi-allsky timers need to be switched out for service control functions for the system units in views.py and the template.

System Info -> System -> Services [tab]

TN-1:
> The service control function for the indiserver and indi-allsky timers need to be switched out for service control functions for the system units in views.py and the template.

Yup good catch, I left the System Info page as it was in my deb branch for backwards compatibility, and when I backported the work I missed that. I will make the change :)

PR #3121, opened 2026-09-05
Improve loop and mini-timelapse preview playback

321MathiasS123:
Normal and panorama loops, including both mini-timelapse previews, currently load each image independently and display it as soon as loading finishes. On slower connections or at higher frame rates, images can arrive out of order, causing playback to jump backward and forward or display several frames in quick succession. In panorama mini-timelapse previews, this can also make the animated pan jump backward. Loop history changes also wait for the scheduled refresh and the current playback pass to finish.

This change:
- Preloads and decodes a bounded four-frame buffer, presenting frames in playback order and holding the current image while the next frame loads.
- Skips failed or timed-out images and prevents late completions from interrupting playback.
- Requests loop data immediately when the timespan changes and restarts playback when the new list arrives.
- Cancels obsolete loop requests and pending frames, including mini-timelapse frames when the selected period or source changes.
- Avoids resizing the loop canvas on every frame when its dimensions have not changed.
- Adds links between the normal and panorama loop players.

FPS controls, Rock behavior in the loop players, mini-timelapse cropping and animated pans, and the existing three-second end pause are preserved.

Validation: all 260 pytest tests pass, including the entry points for 51 JavaScript playback checks (29 loop and 22 mini-timelapse preview checks). The preview checks cover slow and out-of-order loads, decoding, failures, timeouts, selection changes, and panorama crop and pan behavior. Loop playback and history changes were also verified in the browser.

### DIFF
--- a/tests/flask/loop_playback.test.cjs
+++ b/tests/flask/loop_playback.test.cjs
@@ -1,4 +1,5 @@
-// Run with: node --test tests/loop_playback.test.cjs
+// Also run by test_loop_playback.py as part of the pytest suite.
+// Run directly with: node --test tests/flask/loop_playback.test.cjs
 const assert = require('node:assert/strict');
 const fs = require('node:fs');
 const path = require('node:path');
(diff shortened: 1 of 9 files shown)


tests/flask/loop_playback.test.cjs
TN-1, line 1: Have a look at #3116 and consider porting these tests to the pytest framework to align with the other tests.

321MathiasS123:
Added a follow-up commit adapting the loop tests to #3116’s structure: they now live under `tests/flask/` and run through pytest while retaining all 29 JavaScript checks.

Also added links between the normal and panorama loops beside the playback controls. The panorama link appears only when panorama capture is enabled; the return link is always available. Switching preserves the selected timestamp and player variant.

321MathiasS123:
Added a commit for the mini-timelapse previews, those suffer from the same problem.

321MathiasS123:
Included the fix from #3167 and added tests to prevent that from happening again in the future.

Issue #3128, opened 2026-09-07
Bug: VirtualSky Solve button not hidden when user is unauthenticated


TN-1:
When not logged in, the solve button is still shown. Clicking on it results in nothing happening, however it still presents a progress message to the user. Im assuming it should be hidden in this case so only authenticated users can solve.

@jaxzin

321MathiasS123:
I have had a look at #3127 . If the PR gets merged, this will automatically fix the solve button.

PR #3129, opened 2026-09-07
Support tilted cameras and automatic pointing recovery in VirtualSky

321MathiasS123:
VirtualSky assumes a zenith-pointing camera, which limits alignment for tilted installations.

This change:

- Adds camera elevation and pointing azimuth, independently of image rotation.
- Automatically recovers pointing from stars when the regular solve fails or produces a partial fit.
- Shows all eight calibration values for review before saving.
- Makes the expandable controls compact and responsive, with guidance inside the expanded panel.
- Shows Solve and Save only to users allowed to save configuration, fixing #3128.

Existing zenith installations keep working without configuration changes. Accurate solving requires correct camera location and image timestamps.

Includes regression and adversarial tests: **543 tests passed**. Compatible with PR #3116.


321MathiasS123:
Since the original PR, the following improvements have been added:

- **More reliable camera pointing:** The normal solver now fits lens curvature alongside alignment, reducing the tendency to mistake lens distortion for camera tilt. The search supports different fisheye projections and retries alternative solutions when necessary.
- **Reliability checks:** Solutions are checked for estimated uncertainty and consistency across different parts of the image. Poorly constrained solutions are rejected. For nearly vertical cameras, the solver explains when the pointing direction cannot be determined reliably.
- **Catalogue-date correction:** Both the solver and overlay now account for precession between the star catalogue’s reference date and the image date, preventing this difference from affecting the recovered camera orientation.
- **Optional additional overlay correction:** Remaining mismatches can be corrected using measured star positions after a reliable normal solve. This is disabled by default and accepted only when separate validation stars confirm an improvement. It changes only the overlay; image pixels remain untouched.
- **Better mask and image handling:** The overlay respects the configured opaque image-circle mask, and cardinal-direction labels stay inside the visible image. Optional calibration uses detection masks and the configured sky ROI as hints, accounting for binning, rotation, flips, cropping, scaling and borders.
- **Loading and saving fixes:** Fixed the intermittent “No image loaded yet” error when opening VirtualSky from the gallery. Additional correction can now be saved correctly, including cases where no useful correction was found. The enabled preference is retained, and incompatible saved corrections are prevented from being applied.
- **Clearer interface:** Config and VirtualSky use consistent labels. “Camera pointing direction” describes the camera’s lean, while the existing “Azimuth” setting controls overlay rotation. Help explains location/time requirements, the separate image-processing and overlay circles, and everything included in Save.
- **Responsive layout:** Fields align more consistently across screen widths. Each overlay toggle is grouped with its corresponding label toggle, and limiting magnitude has a matching box.
- **Expanded regression coverage:** Tests cover different fields of view, sensor shapes, cropped circles, lens projections, distortions, noise, masks, invalid inputs, saving and browser-side behavior. They run through the shared pytest suite and npm test commands.

Existing zenith-pointing installations require no configuration changes. Previously saved alignments retain their original projection and catalogue convention until solved again.

Validation included 592 simulated camera configurations, with no incorrect successful solves under the tested accuracy criteria; insufficient or unsupported cases were rejected. Six real images from one fixed camera produced a total pointing-direction spread of 0.68°. The latest integration also passed 906 regression checks; the unchanged camera simulation suite had passed previously.

PR #3134, opened 2026-09-09
Feature: Docker Stack Rewrite

TN-1:
Depends on #3064 

This replaces the source-compilation Docker build with a single Bookworm image that installs INDI Allsky via the APT repository. The updated compose stack runs MariaDB 11, Mosquitto 2, and Nginx, while the entrypoint automatically consolidates legacy directory layouts to preserve existing data volumes and .env settings without manual reconfiguration.

- Switches container builds from source compilation to APT packages
- Adds legacy migration directory consolidation and config bootstrap to the entrypoint
- Updates compose stack to MariaDB 11, Mosquitto 2, and Nginx
- Adds a GitHub Actions workflow to publish multi-arch images (linux/amd64,linux/arm64) to GHCR (So end users can pull images instead of needing to build them)
- Fixes CFA_TYPE missing-key errors on monochrome cameras and guards the syslog handler when /dev/log is absent

I have tested and validated that migrations work correctly from the old files to the new. Build times are cut from 20 minutes to 6 minutes on my laptop, and once hosted GHCR images exist users will be able to pull images in seconds.


TN-1:
Rebased onto main to account for the apt repo, this PR is now ready for review and/or merging,

TN-1:
Will resolve conflicts and reopen

TN-1:
@aaronwmorris You dont need to add GITHUB_TOKEN to the repo secrets, it is automatically created by github when the action is started. Just FYI.
  
Also, once this is merged and the image is published to GHCR, ill do a follow up PR to use that image by default instead of building it.

PR #3137, opened 2026-09-10
Fix: FITS viewer for remote and S3 images on web-only instances

### LINKED ISSUE #3135: FITS viewer (preview and download) fails with "file not found" / HTTP 500 when camera uses non-local (S3) image storage
## Environment

- indi-allsky commit: `c257705c2a58e3a3ab75afb71bb5385bebf3f018` (2026-08-26 16:46:24 -0500)
- Branch: `main`
- Config `__config_level__`: `20260724.0`
- OS: Ubuntu (EC2 instance)
- Python: 3.12 (project virtualenv)
- Deployment: two-instance setup
  - **Capture instance**: connected to the camera, saves FITS/images locally, and additionally uploads them to an S3 bucket (`S3UPLOAD.ENABLE=true`, `S3UPLOAD.UPLOAD_FITS=true`).
  - **Web instance** (this one, where the bug happens): does not capture images itself. It only runs the Flask/gunicorn web app. It receives sync notifications from the capture instance via SyncAPI configured with `POST_S3=true` and `EMPTY_FILE=true` (i.e. the sync payload does not contain the actual file bytes — only metadata — because the real file is expected to be fetched from S3 when needed).
  - The camera record on the web instance has `web_nonlocal_images = 1` in the database (this is the documented setting for "images are not local, fetch from S3").
  - The regular image gallery (JPG thumbnails/full images) and the timelapse videos work correctly on the web instance under this same setup — they load directly from the S3 bucket URL in the browser, with no server-side involvement. This confirms the S3 bucket itself, its public-read policy, and the `s3_prefix` stored on the camera record are all correct and working.

## Summary

On the web-only instance described above, the FITS image viewer (`/indi-allsky/fitsimageviewer`) is broken in two distinct ways, even though the same underlying FITS files load and download fine on the capture instance (where they exist on local disk):

1. Selecting a FITS image to preview it (the on-page JPEG conversion) fails with an HTTP 500 error.
2. Clicking "Download FITS" fails — Chrome's download manager reports "El archivo no se encontraba disponible en el sitio" ("The file was not available on the site").

The regular image gallery does not have this problem; only the dedicated FITS viewer is affected.

### 1) Preview (fits2jpeg) — HTTP 500, FileNotFoundError
Image

Browser console shows repeated `500 (Internal Server Error)` responses from `GET /indi-allsky/fits2jpeg?id=...`.

Server-side log (`/var/log/indi-allsky/webapp-indi-allsky.log`) shows the underlying exception for one of these requests:

```
Exception on /indi-allsky/fits2jpeg [GET]
Traceback (most recent call last):
  File ".../flask/app.py", line 1511, in wsgi_app
    response = self.full_dispatch_request()
  ...
  File ".../indi_allsky/flask/views.py", line 5033, in dispatch_request
    hdulist = fits.open(filename_p)
  File ".../astropy/io/fits/hdu/hdulist.py", line 220, in fitsopen
    return HDUList.fromfile(
  ...
  File ".../astropy/io/fits/file.py", line 700, in _open_filename
    self._file = open(self.name, IO_FITS_MODES[mode])
FileNotFoundError: [Errno 2] No such file or directory: '/var/www/html/allsky/images/<camera_uuid>/fits/20260908/night/09_06/ccd1_20260909_063202.fit'
```

The same `FileNotFoundError` pattern (same view, `views.py` line ~5033) repeats for every FITS image tried, always pointing at a local filesystem path that was never populated on this instance (the file only exists in S3 on this instance).

I confirmed the file genuinely exists and is publicly reachable in the S3 bucket, under a slightly different key than the local path that was attempted:

```
$ aws s3 ls s3://<bucket>/ --recursive | grep ccd1_20260909_063202
... ccd_<camera_uuid>/fits/20260909/night/09_06/ccd1_20260909_063202.fit

$ curl -I "https://<bucket>.s3.<region>.amazonaws.com/ccd_<camera_uuid>/fits/20260909/night/09_06/ccd1_20260909_063202.fit"
HTTP/1.1 200 OK
...
```

(Note the date folder differs: the local path attempted `.../fits/20260908/...` while the real S3 object is under `.../fits/20260909/...` — for the same image. I did not investigate why these differ, but it means any fix that tries to guess the S3 key from the local path convention would be unreliable. The `fitsimage` table in the database already stores the correct key in its `s3_key` column, e.g.:)

```
sqlite3 -table -header /var/lib/indi-allsky/indi-allsky.sqlite \
  "SELECT filename, remote_url, s3_key FROM fitsimage WHERE filename LIKE '%ccd1_20260909_063202%';"

filename: /var/www/html/allsky/images/<camera_uuid>/fits/20260908/night/09_06/ccd1_20260909_063202.fit
remote_url: (empty)
s3_key: ccd_<camera_uuid>/fits/20260909/night/09_06/ccd1_20260909_063202.fit
```

### 2) Download FITS — file not found in the browser

Clicking "Download FITS" in the viewer produces a link whose `href` points at a **relative local path**, e.g.:

```
images/<camera_uuid>/fits/20260908/night/09_06/ccd1_20260909_063202.fit
```

Image

Since the file was never synced to local disk on this instance (only to S3), the request for that path 404s.

For comparison, the regular gallery view's "download image" link on this same instance correctly points at the full S3 URL (e.g. `https://<bucket>.s3.<region>.amazonaws.com/ccd_<camera_uuid>/exposures/.../ccd1_20260909_063202.jpg`) and downloads successfully. So the "local vs. S3 URL" decision clearly exists and works elsewhere in the app — it's specifically not being applied for FITS downloads.

## Additional context for maintainers

I traced this as far as I could without being a contributor to the codebase, in case it saves time:

- `indi_allsky/flask/views.py`, `Fits2JpegView.dispatch_request()` (around line 5033) calls `fits.open(filename_p)` where `filename_p` comes directly from `fits_entry.getFilesystemPath()`, with no check for whether the file exists locally and no fallback to fetch it from S3 even though the DB record (`fitsimage.s3_key`) has what's needed to locate it. (I noticed `JsonImageProcessingView.dispatch_request()`, used by the "FITS Processing" button, has the same pattern around line 9598 — I haven't tested that button myself, so I can't confirm it fails in practice, just flagging it in case it's the same root cause.)
- `indi_allsky/flask/forms.py`, class `IndiAllskyFitsImageViewer` (used by the FITS viewer page) never receives or stores `s3_prefix`/`local` in its `__init__`, unlike the class used for the regular gallery (`IndiAllskyImageViewer`), which does receive and use them. Its `getImages()` method builds the download URL with `img.getUrl(local=True)` — hardcoded to `True` — instead of `img.getUrl(s3_prefix=self.s3_prefix, local=self.local)` like the gallery's equivalent code does.
- `indi_allsky/flask/views.py`, in `AjaxFitsImageViewerView.dispatch_request()` and `FitsImageViewerView.get_context()`, the `IndiAllskyFitsImageViewer` / `IndiAllskyFitsImageViewerPreload` objects are instantiated without passing `s3_prefix`/`local`, unlike `GalleryViewerView`, which computes `local` from `self.web_nonlocal_images` / `self.web_local_images_admin` / `self.verify_admin_network()` and passes it through.

So it looks like the FITS viewer simply never received the same "non-local images" support that the regular gallery already has, rather than this being a configuration issue on my end — the JPG gallery and timelapses work correctly under the exact same `web_nonlocal_images=1` setup.

## Note on this instance's own S3UPLOAD config

This web instance's own `S3UPLOAD` section is left at default values (`BUCKET: "change-me"`,
empty credentials) because this instance never uploads to S3 — only the capture instance does.
This is expected and, as far as I can tell, irrelevant to this bug: the actual pointer to the
S3 object is stored per-camera (`camera.s3_prefix`) and per-file (`fitsimage.s3_key`), both of
which are correctly populated (see the `curl` test above returning 200 OK). I'm noting this
just in case it looks suspicious at a glance — happy to paste the redacted section if it's
useful anyway.

Happy to test a patch or provide more logs/config if useful.

TN-1:
- Cache remote FITS files locally when generating JPEG previews in fits2jpeg and image processing
- Return remote/S3 download URLs instead of local paths when non-local
- Filter dates and dropdown choices by remote/S3 availability when local images are disabled
- Use database createDate for timestamp calculations instead of file mtime

Closes #3135 

Draft while awaiting testing by the reporter

### DIFF
--- a/indi_allsky/flask/forms.py
+++ b/indi_allsky/flask/forms.py
@@ -7387,7 +7387,9 @@ class IndiAllskyFitsImageViewer(FlaskForm):
     def __init__(self, *args, **kwargs):
         super(IndiAllskyFitsImageViewer, self).__init__(*args, **kwargs)
 
+        self.s3_prefix = kwargs.get('s3_prefix', '')
         self.camera_id = kwargs.get('camera_id')
+        self.local = kwargs.get('local', True)
 
 
     def getYears(self):
@@ -7395,6 +7397,17 @@ class IndiAllskyFitsImageViewer(FlaskForm):
             self.model.createDate_year,
         )\
             .filter(self.model.camera_id == self.camera_id)
+
+
+        if not self.local:
+            # Do not serve local assets
+            years_query = years_query\
+                .filter(
+                    or_(
+                        self.model.remote_url != sa_null(),
+                        self.model.s3_key != sa_null(),
+                    )
+                )
 
 
         years_query = years_query\
@@ -7423,7 +7436,18 @@ class IndiAllskyFitsImageViewer(FlaskForm):
                     self.model.camera_id == self.camera_id,
                     self.model.createDate_year == year,
                 )
-        )
+            )
+
+
+        if not self.local:
+            # Do not serve local assets
+            months_query = months_query\
+                .filter(
+                    or_(
+                        self.model.remote_url != sa_null(),
+                        self.model.s3_key != sa_null(),
+                    )
+                )
 
 
         months_query = months_query\
@@ -7454,7 +7478,18 @@ class IndiAllskyFitsImageViewer(FlaskForm):
                     self.model.createDate_year == year,
                     self.model.createDate_month == month,
                 )
-        )
+            )
+
+
+        if not self.local:
+            # Do not serve local assets
+            days_query = days_query\
+                .filter(
+                    or_(
+                        self.model.remote_url != sa_null(),
+                        self.model.s3_key != sa_null(),
+                    )
+                )
 
 
         days_query = days_query\
@@ -7485,7 +7520,18 @@ class IndiAllskyFitsImageViewer(FlaskForm):
                     self.model.createDate_month == month,
                     self.model.createDate_day == day,
                 )
-        )
+            )
+
+
+        if not self.local:
+            # Do not serve local assets
+            hours_query = hours_query\
+                .filter(
+                    or_(
+                        self.model.remote_url != sa_null(),
+                        self.model.s3_key != sa_null(),
+                    )
+                )
 
 
         hours_query = hours_query\
@@ -7514,7 +7560,18 @@ class IndiAllskyFitsImageViewer(FlaskForm):
                     self.model.createDate_day == day,
                     self.model.createDate_hour == hour,
                 )
-        )
+            )
+
+
+        if not self.local:
+            # Do not serve local assets
+            images_query = images_query\
+                .filter(
+                    or_(
+                        self.model.remote_url != sa_null(),
+                        self.model.s3_key != sa_null(),
+                    )
+                )
 
 
         images_query = images_query\
@@ -7540,7 +7597,11 @@ class IndiAllskyFitsImageViewer(FlaskForm):
                     role_names,
                 )
 
-            fits_url = img.getUrl(local=True)
+            try:
+                fits_url = img.getUrl(s3_prefix=self.s3_prefix, local=self.local)
+            except ValueError as e:
+                app.logger.error('Error determining relative file name: %s', str(e))
+                continue
 
             image_dict = dict()
             image_dict['id'] = img.id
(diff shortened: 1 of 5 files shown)


TN-1:
Tested and verified. Reaady to merge.

PR #3160, opened 2026-09-15
Fix deb config files

TN-1:
So that change to `service/apache_indi-allsky.conf` was actually intentional, I was still copying them from `service/`, I just misremembered things. This moves copies of the config files to `debian/` so that the deb package can install its own modified files, while setup.sh retains its original files.

@aaronwmorris this needs merging to fix the nightly build


TN-1:
Stable will be ok, it got the files before the revert

PR #3163, opened 2026-09-15
Replace apache with caddy reverse proxy

aaronwmorris:


aaronwmorris:
@TN-1 This is my first pass at using Caddy

TN-1:
@aaronwmorris Do we just want to support caddy alone and remove support for apache and nginx? Simplify things a bit

TN-1:
Instead of rewriting the config file, lets make a complete config thats ready to import in debian/conf/ and then drop that into `/etc/caddy/conf.d/` (Need to create). Then in the caddy file we can prepend `import /etc/caddy/conf.d/*.conf` so local user changes are not overwritten. This follows the same pattern that apache and nginx use.

As an effect of that, the global options block will need to be tweaked like:
```
:%HTTP_PORT%, :%HTTPS_PORT% {
tls internal
   tls {
            on_demand
         }
```

Would be something like:
```
configure_caddy() {
    mkdir -p /etc/caddy/conf.d

    # 1. Generate the site config snippet into conf.d
    sed \
        -e "s|%IMAGE_FOLDER%|$IMAGE_FOLDER|g" \
        -e "s|%HTTP_PORT%|${HTTP_PORT:-80}|g" \
        -e "s|%HTTPS_PORT%|${HTTPS_PORT:-443}|g" \
        /usr/share/indi-allsky/service/caddy_indi-allsky.conf > /etc/caddy/conf.d/indi-allsky.conf
    chmod 644 /etc/caddy/conf.d/indi-allsky.conf

    # 2. Ensure Caddyfile imports conf.d without overwriting user settings
    CADDYFILE="/etc/caddy/Caddyfile"
    IMPORT_LINE="import /etc/caddy/conf.d/*.conf"

    if [ ! -f "$CADDYFILE" ]; then
        echo "$IMPORT_LINE" > "$CADDYFILE"
        chmod 644 "$CADDYFILE"

    elif ! grep -qsF "$IMPORT_LINE" "$CADDYFILE"; then
        TMPFILE=$(mktemp)
        echo "$IMPORT_LINE" > "$TMPFILE"
        cat "$CADDYFILE" >> "$TMPFILE"
        mv "$TMPFILE" "$CADDYFILE"
        chmod 644 "$CADDYFILE"
    fi

    # 3. Reload Caddy if active, or start/enable if not running
    if systemctl is-active --quiet caddy; then
        systemctl reload caddy || systemctl restart caddy
    else
        systemctl enable --now caddy
    fi
}
```

The config file would become:
```
:%HTTP_PORT%, :%HTTPS_PORT% {
    tls internal
    tls {
        on_demand
    }
    redir / /indi-allsky/
    ### indi-allsky config
    handle /indi-allsky* {
        handle /images/* {
            root * %IMAGE_FOLDER%
            file_server
        }
        handle /static/* {
            root * /usr/share/indi-allsky/indi_allsky/flask/static
            file_server
        }
        handle {
            reverse_proxy unix//var/lib/indi-allsky/gunicorn-indi-allsky.sock
        }
    }
    ### ADSB Aircraft tracking
    handle /skyaware* {
        handle /data/* {
            root * /run/dump1090-fa
            file_server
        }
        handle /data-978/* {
            root * /run/skyaware978
            file_server
        }
        handle {
            root * /usr/share/skyaware/html
            file_server
        }
    }
}
```

PR #3166, opened 2026-09-16
Fix: Caddy improvements

TN-1:
@aaronwmorris 

Refactor Caddy setup in the Debian package to drop site snippets into `/etc/caddy/conf.d/indi-allsky.conf` and prepend an import directive to `/etc/caddy/Caddyfile`, preserving user customizations. Also updates the socket path to `/run/indi-allsky/gunicorn-indi-allsky.sock` and grants the caddy user www-data group permissions.


aaronwmorris:
You caught me in the middle of trying to fix the caddy issues.  I am still getting used to the structure of building the packages, so I am having to learn as I go.

I have fixed the deployment issues with #3172 .  I found a few other issues, as well, such as the logging was not working properly.  I took some of what you submitted in the PR and integrated it with my change, so obviously, I will not be merging this request.

The caddy config as you submitted will not work.  I tested the `:80 :443 {` syntax and found it did not work correctly if the ports were altered.  Using `:444` would setup a http [non-https] listener.  I am okay with keeping the other web server configs for now, but we may yank them later.

TN-1:
No worries, I'll have a look through that PR and see if anything screams out

PR #3172, opened 2026-09-16
Apt and Caddy updates

aaronwmorris:


TN-1:
Looks good to me :+1:

TN-1:
Have tested an apache to caddy migration on my prod system, no issues to report.

PR #3201, opened 2026-09-21
Support automatic VirtualSky mirroring

### LINKED ISSUE #3171: VirtualSky - option to reverse/invert axes (Y and X)
In my allsky, I have West set to the left and East to the right, however in the new VirtualSky this is reversed, and I see no way to put it the same way I have the image (which I don't really want to change).

Image

Is there any way to change this?

321MathiasS123:
Adds horizontal and vertical overlay flipping and lets Solve determine mirroring and rotation from star matches, including for tilted cameras. The captured image remains unchanged, and overlay labels stay readable.

Orientation is saved and restored with the alignment. Projection, pointer coordinates and optional overlay correction use the same orientation, while image masks remain in photo coordinates. Both flip settings default to off.

Validation:
- 289 focused pytest regressions passed, including Node browser tests.
- Tested independently flipped real night captures, rotations and simulated camera tilt.
- All 94 successful solves in the real-image exercise recovered the correct handedness; poor or insufficiently constrained frames were refused.

Fixes #3171.


321MathiasS123:
I noticed that the "dot" for Jupiter in VirtualSky was off by almost a degree from the actual position in the sky. This is due to a coordinate wrapping bug in VirtualSky which seems to have caused the overlay position of Jupiter to jump around a lot in 2022 when it crossed the 0°/360° boundary. The bug was still there, though, and is could affect other planets in the future. There was an old commit called "Fudge to fix Jupiter" (https://github.com/slowe/VirtualSky/commit/b85c499c8234fbf506e1bb5fd2ea2cc8ff48fdd0) which mitigated that for Jupiter only at the cost of a large deviation between overlay and actual position.

The newest commit fixes the underlying bug and removes the fudge.

